PR #347, opened 2018-10-22
Prepare for release

bshannon:
Switch to jakarta coordinates.
Configure nexus-staging-maven-plugin.
Use activation snapshot release (for now).

### DIFF
--- a/android/activation/pom.xml
+++ b/android/activation/pom.xml
@@ -39,7 +39,7 @@
     <properties>
 	<activation.spec.version>1.2</activation.spec.version>
 	<activation.extensionName>
-	    javax.activation
+	    jakarta.activation.android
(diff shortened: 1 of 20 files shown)


android/activation/pom.xml
```diff
@@ -39,7 +39,7 @@
     <properties>
 	<activation.spec.version>1.2</activation.spec.version>
 	<activation.extensionName>
-	    javax.activation
+	    jakarta.activation.android
```
vinayvishal: I see apart from replacing `javax` with `jakarta`, `android` too has been appended to the extension name. I hope this is as expected?

bshannon: Yes, I want to distinguish this special Android version from the real JAF version.

PR #353, opened 2018-11-08
Forgot to change JPMS name to jakarta.mail

bshannon:


m0mus:
Why are you changing module names now? I thought that we agreed not changing it before the Eclipse GlassFish release.

bshannon:
The previous release of JavaMail defined a module name, even though it's not in the spec.
The sooner the java.* module name is eliminated, the better.  I'm not going to remove the
module name entirely and undo all the work I did for the previous release.

m0mus:
It's fine to keep java.* module name. You don't need to undo all your previous work. **But you shouldn't change it to jakarta now**. If you do, all modules with javamail dependency must be updated. It's what I'm trying to avoid.

bshannon:
I'm sure the number of applications using JPMS with JavaMail is very small, but the longer a version is out there with the "bad" module name, the more impact it will have when we change it.  I'm changing it now to reduce the impact of the change and avoid the need to create a compatibility module.

PR #396, opened 2019-09-20
Load default providers from mail.jar

MattGill98:
In order to read from the mail.jar instead of the application, a different classloader must be used

### DIFF
--- a/mail/src/main/java/javax/mail/Session.java
+++ b/mail/src/main/java/javax/mail/Session.java
@@ -1,5 +1,6 @@
 /*
- * Copyright (c) 1997, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2020 Oracle and/or its affiliates and others.
+ * All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -12,6 +13,9 @@
  * https://www.gnu.org/software/classpath/license.html.
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ *
+ * Contributors:
+ *   Payara Services - Load default providers from mail.jar classloader
  */
 
 package javax.mail;
@@ -970,6 +974,9 @@ public final class Session {
 	loadAllResources("META-INF/javamail.providers", cl, loader);
 
 	// load default META-INF/javamail.default.providers from mail.jar file
+	loadResource("/META-INF/javamail.default.providers", getClass(), loader, true);
+
+	// Load default META-INF/javamail.default.providers from the the application (not required)
 	loadResource("/META-INF/javamail.default.providers", cl, loader, false);
 
 	// finally, add all the default services
@@ -1098,7 +1105,10 @@ public final class Session {
 	};
 
 	// load default META-INF/javamail.default.address.map from mail.jar
-	loadResource("/META-INF/javamail.default.address.map", cl, loader, true);
+	loadResource("/META-INF/javamail.default.address.map", getClass(), loader, true);
+
+	// Load default META-INF/javamail.default.address.map from the the application (not required)
+	loadResource("/META-INF/javamail.default.address.map", cl, loader, false);
 
 	// load the META-INF/javamail.address.map file supplied by an app
 	loadAllResources("META-INF/javamail.address.map", cl, loader);


MattGill98:
IMO it doesn't matter who's using the broken functionality. The documentation covers this explicitly:

> The `javamail.default.`*X* default files
> are part of the JavaMail `mail.jar` file and should not be
> supplied by users.

If users upgrade the version, it would be reasonable to expect the functionality to change *to* how it's documented.

To be clear this change doesn't implement a deviation from the documentation, but makes the functionality conform as intended. The current functionality of always printing a `WARNING` message for each file clearly labelled `default` is not reasonable.

If you still think this doesn't need merging, I'd like to request the opinion of another committer to help resolve this.

bshannon:
The "should not be supplied by users" comment was only added in JavaMail 1.6.0.
For better or for worse, people have supplied their own default files, and I don't see
any reason to break these applications, and definitely not in a dot-dot release.  There
will be enough other breaking changes coming with the module system that changes
of this sort should be deferred until then.

In what situation are you getting the WARNING message?  Maybe we can fix that in
the short term.

Cousjava:
This occurs on every deployment on Payara.

bshannon:
I don't believe this happens in GlassFish.  What's different about Payara?
What JavaMail jar files are included in Payara?

Cousjava:
it occurs when a new mail session is initialised, and this happens on both GlassFish and Payara (not every deployment, just ones using Mail)

bshannon:
> it occurs when a new mail session is initialised, and this happens on both GlassFish and Payara (not every deployment, just ones using Mail)

I can't reproduce this.  I deployed the JavaMail "webapp" demo program to Eclipse GlassFish 5.1, opened it, and connected to a mail server, and I don't get any warnings in server.log.  How are you reproducing this?

bshannon:
Without a way to reproduce this problem, I'm going to have to close this bug report.

MattGill98:
I've amended the commit to maintain the old behaviour, but also try to load the file from the spec defined location. You've said that GF doesn't show this error, so I'm going to look into why that is

MattGill98:
Glassfish does produce this error, you just need to initialise the SSLSession by triggering an email sending. On initialisation the error is in the logs

bshannon:
Sorry, I've lost track of what warning message you're getting.

Does this only happen when using SSL to connect to the mail server?

MattGill98:
My apologies, I got mixed up with another issue. I've amended my last comment with the correct terminology and the log message

bshannon:
Why is the expected resource not found?  Are you using a mail.jar file that is missing
those resources?

As I said, I deployed the JavaMail sample webapp to GlassFish and connected to a mail
server but never got these error messages in the log file.

MattGill98:
As I say, it occurs on GlassFish 5.1.0 when trying to send an email. For reference I'm using the following class:

~~~
public class JavaMailClient {

    private Session session;

    private String username = "my-email@inbox.com";

    private String password = "mypassword";

    public JavaMailClient() {
        Properties props = new Properties();
        props.put("mail.smtp.host", "smtp.office365.com");
        props.put("mail.smtp.port", "587");
        props.put("mail.smtp.auth", "true");
        props.put("mail.smtp.starttls.enable", "true");

        this.session = getInstance(props, new Authenticator() {
            @SuppressWarnings("PMD.AccessorMethodGeneration")
            @Override
            protected PasswordAuthentication getPasswordAuthentication() {
                return new PasswordAuthentication(username, password);
            }
        });
    }

    public void sendMessageToSelf(String subject, String body) throws AddressException, MessagingException {
        Message message = new MimeMessage(this.session);
        // Set from and to
        message.setFrom(new InternetAddress(username));
        message.setRecipient(TO, new InternetAddress(username));
        // Set content
        message.setSubject(subject);
        message.setContent(body, TEXT_PLAIN);
        // Send
        Transport.send(message);
    }

}
~~~

bshannon:
What triggers the sendMessageToSelf method?

MattGill98:
Whatever you want, I don't think it's too relevant to the reproducer

bshannon:
I'm sorry, I just can't reproduce this.

I've tried my code.

I've tried your code above.

I've used GlassFish 5.1 downloaded from Eclipse and unchanged.

Nothing causes those warnings you're seeing.
Unless you have some other ideas about how to reproduce this,
I'm going to have to close this issue.

MattGill98:
So the full steps:

Download GF 5.1.0 from https://projects.eclipse.org/projects/ee4j.glassfish/downloads

Start the domain:

~~~
bin/asadmin start-domain -v
~~~

Deploy the application to the root context:
~~~
bin/asadmin deploy --contextroot  / javamail-1.0-SNAPSHOT.war
~~~

Trigger the email (it doesn't matter whether it actually sends).

~~~
curl -X POST -d "Hello World" -H "Content-Type: text/plain" http://localhost:8080/api/mail
[javamail-demo.zip](https://github.com/eclipse-ee4j/mail/files/4204106/javamail-demo.zip)
[javamail-demo.zip](https://github.com/eclipse-ee4j/mail/files/4204107/javamail-demo.zip)


~~~

Look in the logs for:

~~~
[#|2020-02-14T10:43:27.561+0000|WARNING|glassfish 5.1|javax.mail|_ThreadID=31;_ThreadName=http-listener-1(1);_TimeMillis=1581677007561;_LevelValue=900;|
  expected resource not found: /META-INF/javamail.default.address.map|#]
~~~

I've attached a project with the source and artifact

cubastanley:
I've just tried this reproducer following the steps detailed above and also get the same error result

bshannon:
Interesting.  I can reproduce it with your sample application.  Let me see if I can figure out
what's different about it.

bshannon:
This appears to be a GlassFish class loader problem, not a problem in JavaMail.

The reason you're getting the warnings is because in fact those resources are not loadable,
even though they should be.  The reason JavaMail works at all is because I added wired in
defaults for the case where these resource files can not be found due to broken class loaders.

So, the fix isn't in JavaMail to suppress the correct warning message.  The fix is somewhere
else in GlassFish to fix class loading.

I need to do a bit more digging to figure out why the class loading is different in my test case
using a Servlet vs. your test case using a JAX-RS class.

PR #411, opened 2020-01-03
BASE64DecoderStream#read: Avoid returning -1 (EOF) in when length is 0

### LINKED ISSUE #408: BASE64DecoderStream.read(buf, off, len) returns EOF when len is 0
**Describe the bug**
The method `com.sun.mail.util.BASE64DecoderStream.read(byte[] buf, int off, int len)` returns `-1` when the given `len` is `0`, indicating EOF.
When `len` is `0`, all loops and conditionals except lines 117,118 are skipped, leading to `off == off0` in line 142 and returning the `-1` although there could be more bytes to read in the InputStream.
This is problematic especially when using `InputStream.readAllBytes()` introduced in JDK 9 to read mail attachments bigger than the `InputStream.DEFAULT_BUFFER_SIZE` of 8192 bytes. The buffer will be read until full, then it tries to read 0 further bytes and the BASE64DecoderStream returns EOF so `readAllBytes()` just returns the first 8192 bytes of the attachment.

**To Reproduce**
```
import java.nio.charset.StandardCharsets;
import java.util.Base64;
import java.util.stream.IntStream;
import static java.util.stream.Collectors.joining;
@Test
void InputStream_readAllBytes() throws IOException {
    byte[] decoded = IntStream.range(0, 10_000).mapToObj(i -> "A").collect(joining()).getBytes(StandardCharsets.UTF_8);
    byte[] encoded = Base64.getEncoder().encode(decoded);
    assertTrue(decoded.length > 8192); // Exceed InputStream.DEFAULT_BUFFER_SIZE
    BASE64DecoderStream sut = new BASE64DecoderStream(new ByteArrayInputStream(encoded));
    // Exercise
    byte[] result = sut.readAllBytes();
    // Verify
    assertArrayEquals(decoded, result);
}
```
Result: `java.lang.AssertionError: array lengths differed, expected.length=10000 actual.length=8192`

**Expected behavior**
The read byte array should equal the input byte array.
The default implementation of `read(byte[] buf, int off, int len)` in `InputStream` always returns `0` when `len` is `0`.

**Tested Java versions:**
-  Java(TM) SE Runtime Environment (build 9.0.4+11)
- OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.5+10)

**Mail server:**
 - Protocol being used: imap
 - Vendor/product: Strato AG
 - Mail service URL: imap.strato.de

**Additional context**
The implementation of `InputStream.readAllBytes()` is correct and works properly e.g. with FileInputStream when files are bigger than 8192 bytes. 
The bug does not occur when `InputStream.transferTo(OutputStream)` is used. As hotfix we use
```
try (final ByteArrayOutputStream bOut = new ByteArrayOutputStream()) {
    BASE64DecoderStream.transferTo(bOut);
    return bOut.toByteArray();
}
```

### LINKED ISSUE #409: API and impl jar should have same module-name
**Describe the bug**
missing module-name/automatic module name in an API jar (thus reliance on the filename based module resolution) makes transition to JDK11/JPMS not so straightforward

**To Reproduce**
* to make sure compilation gets just API and runtime/tests do get also an implementation ('test' scope then needs to explicitly exclude API) on JDK 8 one can use following setup in a maven project:
```
    <dependencies>
        <dependency>
            <groupId>jakarta.mail</groupId>
            <artifactId>jakarta.mail-api</artifactId>
            <version>1.6.4</version>
            <scope>provided</scope>
        </dependency>
        <dependency>
            <groupId>com.sun.mail</groupId>
            <artifactId>jakarta.mail</artifactId>
            <version>1.6.4</version>
            <scope>runtime</scope> <!-- 'test' might be more common here -->
        </dependency>
    </dependencies>
```
* to move to JDK11, let's create module-info in a project having simple ```Hello world``` main class, so it compiles properly as module:
```
module org.errors.bug {
    requires jakarta.mail.api;
}
```
* run the project:
```
ljungman-mac:target lukas$ java -p jakarta.mail-1.6.4.jar:jakarta.activation-1.2.1.jar:mavenproject14-1.0-SNAPSHOT.jar -m org.errors.bug
Error occurred during initialization of boot layer
java.lang.module.FindException: Module jakarta.mail.api not found, required by org.errors.bug
```
* the only way to make this work properly is to have direct compile dependency on the jakarta mail implementation artifact and to have following in the module info:
```
module org.errors.bug {
    requires jakarta.mail;
}
```

**Expected behavior**
definition of the module name of the jakarta mail implementation should allow me to be able to easily switch between API jar/implementation and/or implementation from other vendor (who may not provide APIs directly within his impl jar), so the best would be if the module name of both artifacts is the same

SplotyCode:
Quote from the JavaDocs of InputStream

> <p> If <code>len</code> is zero, then no bytes are read and <code>0</code> is returned;

This fixes #408

Signed-off-by: david <davidscandurra@gmail.com>

### DIFF
--- a/mail/src/main/java/com/sun/mail/util/BASE64DecoderStream.java
+++ b/mail/src/main/java/com/sun/mail/util/BASE64DecoderStream.java
@@ -140,7 +140,7 @@ public class BASE64DecoderStream extends FilterInputStream {
 	}
 
 	if (off == off0)	// haven't returned any data
-	    return -1;
+	    return len == 0 ? 0 : -1;
 	else			// returned some data before hitting EOF
 	    return off - off0;
     }


bshannon:
Sorry, I didin't notice this Pull Request before fixing the issue myself.

Issue #421, opened 2020-01-30
NPE in MailHandler::isLoggable


anthonyvdotbe:
**Describe the bug**
The test `MailHandlerTest::testIsLoggable` fails with an NPE in `MailHandler::isLoggable`

**To Reproduce**
1. `mvn clean verify` (at commit 7ad9fe3999883a0f8a4bb30fab9ddcbf8926838e) with Maven 3.6.2 and Java 11 (Azul Zulu 11.0.6+10)

**Expected behavior**
The project builds successfully.

**Actual behavior**
The following NPE is thrown. From the stack trace, this seems logical: the test passes a `null` record, and `MailHandler::isLoggable` does `record.getLevel()`.

Not sure how this should be resolved: should `MailHandler::isLoggable` be able to deal with `null` records? Or is the `MailHandlerTest::testIsLoggable` test broken?

On another note: at first sight, none of this code was touched recently, so I don't understand why this hasn't been noticed before?

````
java.lang.NullPointerException
        at com.sun.mail.util.logging.MailHandler.isLoggable(MailHandler.java:579)
        at com.sun.mail.util.logging.MailHandlerTest.testLoggable(MailHandlerTest.java:457)
        at com.sun.mail.util.logging.MailHandlerTest.testIsLoggable(MailHandlerTest.java:439)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at ...
````

jmehrens:
This one has been on my list to fix for a while now.  I'll create a patch for this.

bshannon:
@jmehrens if you can get a fix for this in the next few days, I can include it in the 1.6.5
release I'm getting ready to make.  (Likely the last 1.6.x release.)  If it's going to take a
lot longer than that, please let me know.  Thanks.

jmehrens:
@bshannon Pull request was created let me know if you run into any problems.

jmehrens:
@bshannon I updated the commit.  Also I forgot to modify the CHANGES.txt so that is now included in the new commit.

jmehrens:
@bshannon Does this patch need to be ported to the 2.0.0 branch?  If so, is the worked under this ticket or a new ticket?

bshannon:
Once I finish the 1.6.5 release, I'll merge the 2.0.0 branch into master and master
will become the branch where the 2.0.0 release is done.  I'm hoping I won't have
[...]

PR #422, opened 2020-02-01
NPE in MailHandler::isLoggable #421

jmehrens:
This patch aligns with changes in https://bugs.openjdk.java.net/browse/JDK-8233979.  MailHandler.isLoggable is no longer null hostile.  MailHandler.publish now guards against malicious overrides of isLoggable and will report an error to the error manager instead of throwing back to the caller.

I ditched the idea of trying to probe into the super class to determine if it is null hostile as that triggered SpotBugs warnings which were not worth trying to work around.

### DIFF
--- a/mail/src/test/java/com/sun/mail/util/logging/MailHandlerTest.java
+++ b/mail/src/test/java/com/sun/mail/util/logging/MailHandlerTest.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2009, 2019 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2009, 2018 Jason Mehrens. All rights reserved.
+ * Copyright (c) 2009, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2009, 2020 Jason Mehrens. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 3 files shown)


mail/src/test/java/com/sun/mail/util/logging/MailHandlerTest.java
```diff
@@ -618,6 +600,37 @@ public void testPublish() {
...
+    	MailHandler instance = new MailHandler();
+    	InternalErrorManager em = new InternalErrorManager();
+    	instance.setErrorManager(em);
+    	instance.setLevel(Level.ALL);
+    	instance.publish((LogRecord) null);
+    	instance.close();
+    	for (Throwable t : em.exceptions) {
+            dump(t);
+        }
+        assertEquals(true, em.exceptions.isEmpty());
```
bshannon: assertEquals(true, ... ought to be just assertTrue(...

mail/src/test/java/com/sun/mail/util/logging/MailHandlerTest.java
```diff
@@ -450,33 +450,15 @@ private void testLoggable(Level lvl, LogRecord record) {
...
-        try {
-            result = instance.isLoggable(record);
-            mem = new MemoryHandler(new ConsoleHandler(), 100, Level.OFF);
-            mem.setErrorManager(em);
-            mem.setLevel(lvl);
-            expect = mem.isLoggable(record);
-        } catch (RuntimeException mailEx) {
-            try {
-                if (mem != null) {
-                    fail("MemoryHandler threw and exception: " + mailEx);
```
bshannon: threw *an* exception

jmehrens: Correct.  Good thing is that typo is corrected because the code is deleted :)

mail/src/test/java/com/sun/mail/util/logging/MailHandlerTest.java
jmehrens: Yes.  There are multiple places I need to fix this.  I'll fix it in the new test and do a sweep of this in a later change.

mail/src/test/java/com/sun/mail/util/logging/MailHandlerTest.java
bshannon: D'oh!

bshannon:
@jmehrens 
I had to resolve conflicts in CHANGES.txt (kind of annoying but necessary I guess)
and then because of that I couldn't rebase and merge, I needed to squash and merge.
I think it all came out ok anyway, but let me know if you see a problem.  Because of
the conflict merge, you'll need to be careful how you merge with upstream before
doing any more work in your copy of the repo.

jmehrens:
@bshannon Merge looks good.  I'll work on getting my merges to work more smoothly.  Thanks for the help!

bshannon:
@jmehrens I don't think there's anything you can do to help with merging CHANGES.txt.
If both of us are working on bug fixes, conflict is inevitable.  It's not a big deal to do the
merge in the GitHub GUI.  I think the "second merge" that it complained about was just
a consequence of having to do the first merge.

Bottom line, don't worry about it.  Thanks for the fix!

PR #428, opened 2020-03-12
solve the bcs table index use inconsistency

dslztx:
solve the Bad Character Shift table index use
inconsistency between the 670th line and
823th line which leads to some problem in
non-ascii situation

    Signed-off-by: dslztx <dslztx@gmail.com>


bshannon:
Interesting.  This looks like it should be correct, but do you have a test case that shows the problem?  What happens when it fails?  Is it just slower, or does it produce incorrect
results?

Also, I just have to ask...  is it purely coincidence that you submitted this just days after
I published the latest Jakarta Mail release?  Or did you see the release and that inspired
you to submit this?

In any event, it will likely be quite some time before I'll be able to produce a release that
includes this fix.

dslztx:
I have added a test case which shows that the right result can not be parsed without the bug fix.

For your confusion, I found the bug in 2018 ant I fixed it myself. We maintain our own version for JavaMail jar, So we don't have a lot of motivations to send a feedback to community. But we must change to use the latest JavaMail version in the near future for some reason, so it is just a coincidence.

All depends on you that whether to fix the bug or not and when to.

Thanks for your reply.

bshannon:
Thanks for the test case!

Does the problem only occur when non-ASCII characters are used in the multipart boundary?
Of course, that's a violation of the spec.  Is there a mailer that purposely uses non-ASCII
characters in the boundary?

Still, I agree that this should be fixed, although it's a less important problem in practice.

Also, was there a reason you needed to add the mailcap file to the test resources?  Wasn't
the mailcap file in the source code being found?

dslztx:
Yes, the problem only occur when non-ASCII characters are used in the multipart boundary. And obviously it violates the spec.

We do some work about anti-spam-mail, the spammer take all means including sending non-standard mail for bypassing anti-spam mechanism. We have to provide non-standard mail parsing ability to cope with these situations.

I have deleted the unnecessary file added for test. I am so sorry for unnecessary file addition last commit, although the test case can not be run successfully yet in IntelliJ IDEA without the file , but I find it succeeds using `mvn test` command without the file , therefore I suppose it is just something about IntelliJ IDEA.

bshannon:
Looks like your latest commit is missing the signed-off header.
Yes, I know it seem ridiculous since you're just removing a file that should
never have been added, but please use commit -s ---amend and push -f
so I can merge this.

dslztx:
It is an unintentional commit and I have made an update

lukasj:
closing this one in favor of #475

Issue #403, opened 2019-10-16
Android SocketChannel workaround not working around


ghost:
I am using the android-mail 1.6.4 library with activation.

Regarding the section of code Protocol.java:541 (getChannel) and the workaround:

``` java
// XXX - Android is broken and SSL wrapped sockets don't delegate
// the getChannel method to the wrapped Socket
if (socket instanceof SSLSocket) {
   try {
Field f = socket.getClass().getDeclaredField("socket");
f.setAccessible(true);
Socket s = (Socket)f.get(socket);
ret = s.getChannel();
   } catch (Exception ex) {
// ignore anything that might go wrong
   }
}
return ret;
    }
```

socket.getClass().getDeclaredField("socket")
This line returns null and throws the exception:

```
No field socket in class Lcom/android/org/conscrypt/Java8FileDescriptorSocket; (declaration of 'com.android.org.conscrypt.Java8FileDescriptorSocket' appears in /system/framework/conscrypt.jar)
```

I am not sure if this is an issue with a change in android versions, but I am using Android 9 and jdk8 here.

My session and property definitions are:

``` kotlin
    @Provides
    @JvmStatic
    @Singleton
    fun providesMailSession(properties: Properties): Session = Session.getInstance(properties)

    @Provides
    @JvmStatic
    @Singleton
    fun providesMailProperties(): Properties = Properties().apply {
        this["mail.smtp.host"] = "mail.newroztelecom.com"
        this["mail.smtp.port"] = 587
        this["mail.smtp.auth"] = true
        this["mail.smtp.starttls.enable"] = true
        this["mail.imap.host"] = "mail.newroztelecom.com"
        this["mail.imap.port"] = 143
        this["mail.imap.starttls.enable"] = true
        this["mail.imap.usesocketchannels"] = true
    }

    @Provides
    @JvmStatic
    @Singleton
    fun providesIdleManager(es: ExecutorService, properties: Properties) =
        IdleManager(Session.getInstance(properties), es)

    @Provides
    @JvmStatic
    @Singleton
    fun providesMailExecutor(): ExecutorService = Executors.newCachedThreadPool()
```

The IMAPStore is initialized with this session, and connected.

I read that SocketChannels do not support read and write operations so I open the folder with 

``` kotlin
folder.open(javax.mail.Folder.READ_ONLY)
```

And then 
``` kotlin
idleManager.watch(task.folder)
```
throws the exception that the folder is not using SocketChannels  because of the socket being null.

[...]

PR #441, opened 2020-04-02
Android SocketChannel workaround not working around #403

jmehrens:
@bshannon I created a patch for this issue but at the moment I'm unable to build a test for it due to problems with my IDE setup.  I broke out the search logic into a private method so I can invoke it via reflection from the JUnit test.  Might take me a bit to create the test but wanted to at least give you a patch.

### DIFF
--- a/mail/src/main/java/com/sun/mail/iap/Protocol.java
+++ b/mail/src/main/java/com/sun/mail/iap/Protocol.java
@@ -536,21 +536,61 @@ public class Protocol {
     public SocketChannel getChannel() {
 	SocketChannel ret = socket.getChannel();
 	if (ret != null)
-	    return ret;
+		return ret;
 
-	// XXX - Android is broken and SSL wrapped sockets don't delegate
-	// the getChannel method to the wrapped Socket
 	if (socket instanceof SSLSocket) {
-	    try {
-		Field f = socket.getClass().getDeclaredField("socket");
-		f.setAccessible(true);
-		Socket s = (Socket)f.get(socket);
-		ret = s.getChannel();
-	    } catch (Exception ex) {
-		// ignore anything that might go wrong
-	    }
+		ret = Protocol.findSocketChannel(socket);
 	}
 	return ret;
+    }
+
+    /**
+     * Android is broken and SSL wrapped sockets don't delegate
+     * the getChannel method to the wrapped Socket.
+     * 
+     * @param socket a non null socket
+     * @return the SocketChannel or null if not found
+     */
+    private static SocketChannel findSocketChannel(Socket socket) {
+	//Search class hierarchy for field name socket regardless of modifier.
+	for (Class<?> k = socket.getClass(); k != Object.class; k = k.getSuperclass()) {
+		try {
+			Field f = k.getDeclaredField("socket");
+			f.setAccessible(true);
+			Socket s = (Socket)f.get(socket);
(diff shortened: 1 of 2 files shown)


bshannon:
Thanks, @jmehrens!  Do you have an environment in which to test the fix, e.g., using Android Studio?  Have you already verified that the bug exists in the latest version of Android?

jmehrens:
@bshannon I do not have an environment to prove this exists in latest version of Android.
That said, I can prove a few things around this patch.
1. The original workaround code (JakartaMail master) fails if the 'socket' field is buried deep in a super class.  I've tested this with a non JUnit test case by just extending Socket 5 layers deep.
2. Android switched to using Conscrypt and the socket field is buried deep in a super class.
3. Using the patch I proposed I'm able to find the socket field even if it is buried deep in a super class. I've tested this with a non JUnit test case.
4. Using the patch I proposed I'm able to find any references that are declared to hold a socket but are not named 'socket'.  I've tested this with a non Junit test case.

The only case I haven't proven is that running under Android that the SSLSocket is delegating to the 'socket' field in the AbstractConscryptSocket.java.  If it doesn't then it seems like any additional work should be done to make Conscrypt implement SocketChannel correctly.

In summary, this patch is worthwhile addition.  It should at least make JavaMail resilient to changes in the socket class hierarchy which is what happened when Conscrypt was integrated into Android.

I'll keep working to port my non-JUnit test to a working JUnit test.  I'll take a look at Android Studio but it is completely new to me.  So it will take me some time.

jmehrens:
@bshannon I added a commit to include changes for a proper JUnit test.  One thing that might get tricky is FakeSocketChannel and or use -Werror over multiple JDK versions.  It seem like new abstract methods are added to that class every major release of the JDK.  So it going to get complicated with -Werror and Override annotation.  Might need to resort to using a lot of SuppressWarnings("override") to deal with differences in the JDKs.  This might be a candidate for mocking framework.

jmehrens:
@bshannon The more I thought about it it seems the best action would be to get rid of the FakeSocketChannel implementation in my test to avoid excessive maintenance on this test.  Pushed a new and hopefully final change for this issue.

jmehrens:
@lukasj Bill was considering this patch for JakartaMail 2.0.  I'm fine with including it in 2.0 or 2.1.  Either release would be safe.  I don't want to merge this myself without a review as I'm not sure which release to target.

lukasj (approved):
LGTM, thanks! And sorry for delay.

Do you think this is worth to be backported to 1.6.x?

jmehrens:
@lukasj I think it would make sense to backport to 1.6.x.  The IdleManager just broken without this fix under Android. See: https://stackoverflow.com/questions/47346936/javamail-exception-javax-mail-messagingexception-folder-is-not-using-socketcha/

I'll file a ticket with Conscrypt and link it to this issue.  It would be nice if SocketChannel views were supported that codebase too.

I can prepare a PR for the 1.6.x backport over the weekend.

lukasj:
OK, go ahead and prepare it, please. Target the PR to `v1.x` branch. Also do not forget to update copyright headers and changelog (see 4b4b44b37aa348365edcf80e54ae5a04f00a2e45). I used `git cherry-pick...`  for my backports in #497 and I was surprised that almost all changes I did in changelog were applied properly to fall under 1.6.6 entry there :-)

PR #469, opened 2020-08-17
Remove fallback to system default charset solving #467

### LINKED ISSUE #467: Multisegment headers are may use wrong encoding
**Describe the bug**
Multisegment headers (e.g. the (long) filename of an attachment) will be read wrong if:

- the OS' default encoding is not ASCII based AND
- the system property `mail.mime.charset` is not set.

In case of the attachments, the filename will then contain anything but the actual filename.

**To Reproduce**
1. Use an OS with a default encoding set to something non-ASCII based. (e.g. z/OS with IBM-1047 as encoding)
2. Make sure the system property `mail.mime.charset` is **NOT** set.
3. Create yourself a MIME message with an attachment. The attachment needs a long filename that is stored in a multi-segement header
4. Read a MIME message using this mail library (version 1.6.5) and parse the MIME message. 
5. Retrieve the filename of the attachment 
6. See error

**Expected behavior**
The actual filename of the attachment should be returned.

**Environments**
_It's working on_  any system that uses a default encoding based on ASCII. Most Windows machines will work.
_It's NOT working on_ any system using an encoding that is not based on ASCII will not work. For example the IBM z/OS or IBM AIX machines using IBM-1047 or IBM-037 as their default encoding will produce a wrong filename.

**Environment where error has been observed**
 - OS: IBM Z14 Mainframe using z/OS; Java application running in USS (Unix sub system)
 - JVM: `JRE 1.8.0 z/OS s390x-64-Bit Compressed References 20200402_443261 (JIT enabled, AOT enabled) OpenJ9 - 35ef566 OMR - 4bca4f4 IBM - 55acf4a`
 - Using [SubethaSMTP](https://github.com/voodoodyne/subethasmtp) to retreive mails via SMTP
 - Using `com.sun.mail:jakarta.mail:jar:1.6.5` to read the MIME message
 - Using custom application to interpret attachments

noxone:
The MIME message is based on ASCII character set. Thus the default fallback encoding used to read MIME messages should be ASCII or at least based on ASCII.

Using the OS' default encoding may result in using a completely different encoding. Most IBM machines (z-Series and AIX machines) use EBCDIC based encodings as default encoding und thus reading the messages fails.

Closes #467

Signed-off-by: Olaf Neumann <eclipse@olafneumann.org>

### DIFF
--- a/mail/src/main/java/jakarta/mail/internet/MimeUtility.java
+++ b/mail/src/main/java/jakarta/mail/internet/MimeUtility.java
@@ -1291,20 +1291,14 @@ public class MimeUtility {
 	    try {
 		defaultJavaCharset = System.getProperty("file.encoding", 
 							"8859_1");
-	    } catch (SecurityException sex) {
-		
-		class NullInputStream extends InputStream {
-		    @Override
-		    public int read() {
-			return 0;
-		    }
+		} catch (final SecurityException sex) {
+			// fall back to ISO-Latin-1
+			// don't use actual system encoding, because this might be
+			// something completely different, like EBCDIC (IBM-037)
+			if (defaultJavaCharset == null) {
+				defaultJavaCharset = "8859_1";
+			}
 		}
-		InputStreamReader reader = 
-			new InputStreamReader(new NullInputStream());
-		defaultJavaCharset = reader.getEncoding();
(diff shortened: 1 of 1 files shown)


lukasj (approved):
LGTM

PR #475, opened 2020-08-31
solve the bcs table index use inconsistency

dslztx:
solve the Bad Character Shift table index use
inconsistency between the 670th line and
823th line which leads to some problem in
non-ascii situation

And add a test for the solution

    Signed-off-by: dslztx <dslztx@gmail.com>


dslztx:
It is same as [pull request 428](https://github.com/eclipse-ee4j/mail/pull/428), I just rearrange the commit and recreate the pull request.

Cousjava (approved):

lukasj:
@Cousjava how are we going to merge 2.0.0 release branch to master without corrupting history now? We can't respin 2.0.0 now since the ballot and related things are already running.

PR #476, opened 2020-09-03
Added back missing null check

### LINKED ISSUE #456: MimeMessage.setFrom(null) fails instead of removing the header
Even though the method's java doc states:

> Set the RFC 822 "From" header field. Any existing values are replaced with the given address.
> **If address is <code>null</code>, this header is removed.**

The change in ac77368ae15f4a0092499a4e939fe58c21325f37 is deferring the null check to remove the header to `setAddressHeader()`; however it is providing an array with a single null element, which is not what the code expects.

`setFrom(InternetAddress)` fails with the following stacktrace:
```
java.lang.NullPointerException
	at javax.mail.internet.InternetAddress.toUnicodeString(InternetAddress.java:534)
	at javax.mail.internet.MimeMessage.setAddressHeader(MimeMessage.java:731)
	at javax.mail.internet.MimeMessage.setFrom(MimeMessage.java:382)

```
`setFrom(String)` fails with the following stacktrace:
```
java.lang.NullPointerException
	at javax.mail.internet.InternetAddress.parse(InternetAddress.java:769)
	at javax.mail.internet.InternetAddress.parse(InternetAddress.java:728)
	at javax.mail.internet.InternetAddress.parse(InternetAddress.java:705)
	at javax.mail.internet.MimeMessage.setFrom(MimeMessage.java:400)
```

Either the if/else check on each setXxx() method should be reinstated, or setAddressHeader() should expect an array with a single null value.

tweimer:
I have just restored the null checks from ac77368ae15f4a0092499a4e939fe58c21325f37, as it is described in the javadocs:

> If address is null, this header is removed.

Closes #456

### DIFF
--- a/mail/src/main/java/jakarta/mail/internet/MimeMessage.java
+++ b/mail/src/main/java/jakarta/mail/internet/MimeMessage.java
@@ -379,7 +379,10 @@ public class MimeMessage extends Message implements MimePart {
      */
     @Override
     public void setFrom(Address address) throws MessagingException {
-	setAddressHeader("From", new Address[] { address });
+        if (address == null)
+            removeHeader("From");
+        else
+            setAddressHeader("From", new Address[] { address });
     }
 
     /**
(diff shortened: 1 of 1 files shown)


lukasj (approved):
LGTM, I'll add a test for this in the other PR. Thanks!

PR #477, opened 2020-09-04
Issue 461: Added two line authentication support for XOAUTH2 connections

Nubzor:
This is a fix (or rather a feature) for the [issue raised by Craig](https://github.com/eclipse-ee4j/mail/issues/461) related to two line authentication format in XOAUTH2.

**Problem description**:

Currently, Jakarta uses one line authentication format to perform authentication to mail servers which seems to be insufficient as there are providers which use two line authentication format. As of the original issue the functionality was introduced by [Bill and **tested on Gmail** ](https://github.com/eclipse-ee4j/mail/issues/416#issuecomment-591233526) which uses one line authentication format what is clearly described in the [official documentation.](https://developers.google.com/gmail/imap/xoauth2-protocol) (please search for: "Things to note about the POP protocol exchange")

Unfortunately, the [RFC](https://tools.ietf.org/html/rfc1734) doesn't clearly specify what the authorization command should look like, even the given example presents two line authentication format. That implies that the mail servers providers may take different approaches during the implementation. As an example, there is a [Microsoft who recently announced](https://techcommunity.microsoft.com/t5/exchange-team-blog/announcing-oauth-support-for-pop-in-exchange-online/ba-p/1406600) support for OAuth 2 in POP3 who uses a different approach than Google and have implemented [two line authentication format](https://docs.microsoft.com/en-us/exchange/client-developer/legacy-protocols/how-to-authenticate-an-imap-pop-smtp-application-by-using-oauth). (please search for: "command split into two lines in the following")

Having that said, in this form, two line authentication doesn't work with Jakarta mail. 

**Solution description**: 
I have considered a few different approaches while implementing the functionality such as: 
* Fallback to two line authentication if one line authentication fails
    * in the end I dropped this idea as it might have a performance impact
* Implementing another Authenticator and relay on existing property `mail.pop3s.auth.mechanisms`
    * unfortunately, given architecture doesn't allow to do this without major modifications as the property is being used to check server capabilities 
* **In the end I decided to extend the existing OAuth authenticator and introduce another property which has to bet set in order to perform two line authentication** 

The change itself is backward compatible and has been tested with our environment and seems to work as expected. All tests are passing. 

**Please be informed that the branch has been created from 4c092797 revision which is the last commit of the 1.6.x stream. I couldn't do a PR to the 1.6.x stream as the branch seems not to be cut. For us, it would be crucial to have that fix released in the next 1.6.x bugfix release, as we are not ready to migrate to 2.x version. Probably the code is compatible with 2.x version as well though I haven't tested that.**

I would be more than happy to discuss the solution or mentioned problem in case of any doubts.


Nubzor:
@m0mus, @Cousjava - I allowed myself to mention you here as you were accepting some of the previous PR. I want to make sure that you are aware of this PR and would like to ask if someone from the team have could have look at this? 

Thanks!

vijay04991:
Where can I download the jar file with this feature??

Nubzor:
@vijay04991 I am afraid you can't. This still has not been merged, you can download the source code from the feature branch and build the jar yourself.

lukasj (commented):
LGTM, thanks!

lukasj:
It's backported to upcoming 1.6.6 by #502 - note that I slightly changed the property name to follow conventions and it has also been documented (what is not documented, does not exist). Feel free to update the wording in the docs through another PR.

Thanks!

tudvari:
Unfortunately version 1.6.6 isn't a working version please see: https://github.com/eclipse-ee4j/mail/issues/542

lukasj:
1.6.7 is out for some time already

Issue #512, opened 2021-01-26
Improve code coverage in logging-mailhandler tests


jmehrens:
The MailHandlerTest fails due to the bad-host-name being resolved on my ISP.
1. Need to update the test to compute a host name that can't be resolved.
2. Update the code coverage in com.sun.mail.util.logging package after testing with [JaCoCo](https://www.eclemma.org/jacoco/).  This was going to be committed but got lost in the repo moves.
3. Update MailHandler tests to include a host name entry in the session to [prevent host name lookups](https://github.com/eclipse-ee4j/mail/blob/4627f4e7677aeb6ba2242331f3f6ce39c8d6f853/mail/src/main/java/jakarta/mail/internet/InternetAddress.java#L649). This will help speed up the test by avoid DNS lookups.

jmehrens:
Committed to master as: https://github.com/eclipse-ee4j/mail/commit/226d2d6f1db283042537fa9a30dfab798649fea2

I'll add close this issue when I complete the backport.

jmehrens:
Committed to v1.x as https://github.com/eclipse-ee4j/mail/commit/fed14a611a2cffdb92e11ad9f4bfa0512733eb27

PR #518, opened 2021-02-09
Improve code coverage in logging-mailhandler tests #512

jmehrens:
Pull request for issue: https://github.com/eclipse-ee4j/mail/issues/512

I'll backport this to 1x when this is approved.  The details are covered in the issue but the main reason to backport is to deal with finding an unknown host and depending on the network, the bad-host-name can be resolved.

### DIFF
--- a/mail/src/test/java/com/sun/mail/util/logging/MailHandlerTest.java
+++ b/mail/src/test/java/com/sun/mail/util/logging/MailHandlerTest.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2009, 2020 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2009, 2020 Jason Mehrens. All rights reserved.
+ * Copyright (c) 2009, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2009, 2021 Jason Mehrens. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -62,7 +62,8 @@ public class MailHandlerTest extends AbstractLogging {
     /**
      * A host name that can not be resolved.
      */
-    private static final String UNKNOWN_HOST = "bad-host-name";
+    private static volatile String UNKNOWN_HOST;
+
     /**
      * Stores a writable directory that is in the class path and visible to the
      * context class loader.
@@ -77,7 +78,7 @@ public class MailHandlerTest extends AbstractLogging {
     public static void setUpClass() throws Exception {
         checkJVMOptions();
         OPEN_PORT = findOpenPort();
-        checkUnknownHost();
+        UNKNOWN_HOST = findUnknownHost();
         assertTrue(findClassPathDir().isDirectory());
     }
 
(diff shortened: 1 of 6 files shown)


jmehrens:
@lukasj Not sure what is going on with the build error but it shows:
`[WARNING] Rule 1: org.apache.maven.plugins.enforcer.RequireJavaVersion failed with message:
Detected JDK Version: 1.8.0-191 is not in the allowed range [11,).`

I didn't change anything on the pom.xml so I wouldn't think that this PR is the root cause.  What is the expected minimum version that is allowed to run JakartaMail?

lukasj:
the job uses JDK 8 instead of 11 which is required for build now. I'll ask someone to fix the job to use correct JDK. Thanks for heads up!

lukasj (approved):
in general - LGTM, only comment about reflection code - probably something to be addressed in a separate PR (if ever)

mail/src/test/java/com/sun/mail/util/logging/MailHandlerTest.java
```diff
@@ -475,6 +477,96 @@ private void testLoggable(Level lvl, LogRecord record) {
...
+        		if (t == Object.class) {
+        			continue;
+        		}
+
+        		if (t.isAssignableFrom(k)) {
+        			//Check for static field holding declared class.
+        			if (Modifier.isStatic(f.getModifiers())
+        					&& Modifier.isFinal(f.getModifiers())) {
+        				f.setAccessible(true);
+        				Object o = f.get((Object) null);
```
lukasj, line 517: I'm thinking about starting using `MethodHandles` in tests to avoid possible access issues going forward, especially when one needs to call `setAccessible`. Something like: `MethodHandles.privateLookupIn(MailHandler.class, MethodHandles.lookup()).unreflectVarHandle(f).get()` should be valid replacement but that would tight tests to JDK 9+, so not sure that is good idea. What do you think?

jmehrens: Is the minimum version for 1.x JDK7?  I was thinking of backporting this to 1.x and would like to avoid diverging the code at this time if 1x requires a min version less than JDK9.  There are a handful of tests where I call setAccessible(true) I would prefer to tackle them in a follow on PR.  That would also give me time to test with your suggested code.

The alternatives tests to accomplish the intended goal would be:
1. The only other way I can think to avoid  create a tests that [launches a new VM](https://stackoverflow.com/questions/23686948/loading-junit-tests-to-different-jvms) with verbose class loading turned on via MBeans.  Then scrape the output to ensure all the declared classes are loaded.
2. Relax the static field to default access (package protected) and inspect it.  I rather it be private but this is possible.  There are few more places that need promoted access if I go down this road.

So if you are good with it lets create a follow on PR to tackle use of setAccessible(true);

lukasj: > Is the minimum version for 1.x JDK7?

Yes, good point and should stay there forever... multi-release jars can be an option but I have no idea how they work in/for tests as typically one runs tests against a folder with test classes, not against a jar (and that complicates things)

It does not make sense to me to "rewrite" all tests now. Probably better to apply new approach in areas, which we don't want to back port.

PR #522, opened 2021-02-11
Improve code coverage in logging-mailhandler tests #512

jmehrens:
Backport of https://github.com/eclipse-ee4j/mail/issues/512

(cherry picked from commit 226d2d6f1db283042537fa9a30dfab798649fea2)

Conflicts:
	mail/src/test/java/com/sun/mail/util/logging/CompactFormatterTest.java
	mail/src/test/java/com/sun/mail/util/logging/MailHandlerTest.java

### DIFF
--- a/doc/release/CHANGES.txt
+++ b/doc/release/CHANGES.txt
@@ -32,6 +32,7 @@ E  473	Several modules are not included in build when JDK11 is used
 E  489	mime handling crashes on Android 11 debug build
 E  493	javamail.providers file missing from provider jars after 1.6.5
 E  505	WriteTimeoutSocket::getFileDescriptor$ support for Conscrypt
+E  512	Improve code coverage in logging-mailhandler tests
 
 
 		  CHANGES IN THE 1.6.5 RELEASE
--- a/mail/src/test/java/com/sun/mail/util/logging/CollectorFormatterTest.java
+++ b/mail/src/test/java/com/sun/mail/util/logging/CollectorFormatterTest.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2013, 2019 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2013, 2018 Jason Mehrens. All rights reserved.
+ * Copyright (c) 2013, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2021 Jason Mehrens. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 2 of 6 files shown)


jmehrens:
@lukasj The merge conflict in CompactFormatterTest was just copyright years.  The merge conflict in MailHandlerTest was between NonDiscriminatingFormatter1 and StaticInitReAuthenticator.  Git was confused due to package name changes in the javax.mail vs. jakarta.mail.  Those are the areas to double check.

lukasj:
Is this the last thing for 1.6.6 you want or do you have something more in the queue? I’d like to cut it off tomorrow or early next week...

jmehrens:
@lukasj I think I've included the main things I wanted to get into 1.6.6.  Is 2.x also near a freeze?  I'm working on a roadmap of features for logging-mailhandler but I've yet to determine which releases I'm targeting so it is no issue if 2.x is also near a freeze.

jmehrens:
@lukasj Looks like v1.x is failing with:
`[WARNING] Rule 1: org.apache.maven.plugins.enforcer.RequireJavaVersion failed with message:
Detected JDK Version: 1.8.0-191 is not in the allowed range [11,).`  I would assume that JDK8 would be allowed for that branch.

lukasj:
no, both branches do require JDK 11 for build now (there are proper JPMS descriptors now), the 1x has set 7 as the target release to keep the 1.7 compatibility. 

as for 2.x:
* I need 2.0.1 (for EE 9.1) in GlassFish by the end of Feb, but I'd rather have it sooner as I need to integrate it to other projects, like Metro or EclipseLink (which should be in GF by the same date) requiring bit more work
* 2.1/3.0 (or another major version) - this should target EE 9.2/10 where the intent is to release it by the end of this year. What that exactly means is still TBD. Due to that we may want to iterate faster and deliver ie version 4.2.6 (an example :-)) to Jakarta EE.Next

Anyway those are two main targets on the table. If we need anything else (ie 2.0.2) in the meantime, it's fine, we can do that should there be a need for it. I'd only limit the scope of work on 1.x versions.

jmehrens:
@lukasj Make sense.  I think the PR build is failing now with: `0:54:03,521 |-ERROR in ch.qos.logback.core.FileAppender[LOG_FILE] - openFile(/logs/james.log,true) call failed. java.io.FileNotFoundException: /logs/james.log (Permission denied)`.

I do think that you'll want to get this PR into 1.x before you freeze it.

lukasj (approved):
LGTM

PR #534, opened 2021-03-08
Integrate activation 2.0.1

lukasj:
Signed-off-by: Lukas Jungmann <lukas.jungmann@oracle.com>


lukasj:
@jmehrens FYI you may need to run builds with "-Pstaging" option temporarily as JAF 2.0.1 has not been released yet. It's available in staging only. I only want to prepare and do final mail 2.0.1  testing...

PR #539, opened 2021-03-11
Support native in Ubuntu/Debian/Mint

jbescos:
Signed-off-by: Jorge Bescos Gascon <jorge.bescos.gascon@oracle.com>

### DIFF
--- a/mbox/native/pom.xml
+++ b/mbox/native/pom.xml
@@ -1,39 +1,135 @@
-tensions>
-                <configuration>
-                    <compilerProvider>generic</compilerProvider>
-                    <compilerExecutable>${compiler.name}</compilerExecutable>
-                    <linkerExecutable>${linker.name}</linkerExecutable>
-                    <compilerStartOptions>
-                        <compilerStartOption>
-                            ${compiler.start.options}
-                        </compilerStartOption>
-                    </compilerStartOptions>
-                    <linkerStartOptions>
-                        <linkerStartOption>
-                            ${linker.start.options}
-                        </linkerStartOption>
-                    </linkerStartOptions>
-                    <linkerEndOptions>
-                        <linkerEndOption>
-                            ${linker.end.options}
-                        </linkerEndOption>
-                    </linkerEndOptions>
-                    <sources>
-                        <source>
-                            <directory>../src/main/cpp</directory>
-                            <includes>
-                                <include>**/*.c</include>
-                            </includes>
-                        </source>
-                        <source>
-                            <directory>../target/native</directory>
-                            <dependencyAnalysisParticipation>
-                                false
-                            </dependencyAnalysisParticipation>
-                        </source>
-                    </sources>
-                </configuration>
-            </plugin>
+<?xml version="1.0" encoding="iso-8859-1"?>
+<!--
+
+    Copyright (c) 1997, 2021 Oracle and/or its affiliates. All rights reserved.
+
+    This program and the accompanying materials are made available under the
+    terms of the Eclipse Public License v. 2.0, which is available at
+    http://www.eclipse.org/legal/epl-2.0.
+
+    This Source Code may also be made available under the following Secondary
+    Licenses when the conditions for such availability set forth in the
+    Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+    version 2 with the GNU Classpath Exception, which is available at
+    https://www.gnu.org/software/classpath/license.html.
+
+    SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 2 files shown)


mbox/native/pom.xml
```diff
@@ -1,39 +1,178 @@
...
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+			    http://maven.apache.org/maven-v4_0_0.xsd">
+    <parent>
+	<groupId>com.sun.mail</groupId>
+	<artifactId>all</artifactId>
+	<version>2.0.0-RC4</version>
```
lukasj, line 27: should be some current snapshot version

jbescos: Done

lukasj (approved):
LGTM

lukasj (commented):
can you also put a note into [changes.txt](https://github.com/eclipse-ee4j/mail/blob/master/doc/release/CHANGES.txt), please?

jbescos:
I have updated the changes.txt

jbescos:
I need to update a pom.xml because of comments here:
https://github.com/eclipse-ee4j/mail/pull/540

jbescos:
I have added in the native pom.xml the restriction to run it with JDK 1.7 or 8. So there is not need to specify the profile.

More than JDK 8 is not allowed because there is no javah.

pzygielo:
> More than JDK 8 is not allowed because there is no javah.

use something like
```xml
            <plugin>
                <artifactId>maven-compiler-plugin</artifactId>
                <executions>
                    <execution>
                        <id>generate-headers</id>
                        <phase>generate-sources</phase>
                        <goals>
                            <goal>compile</goal>
                        </goals>
                    </execution>
                </executions>
                <configuration>
                    <compilerArgs>
                         <arg>-h</arg><arg>${project.build.directory}/javah</arg>
                    </compilerArgs>
                </configuration>
            </plugin>
```
instead

mbox/native/pom.xml
```diff
@@ -1,39 +1,180 @@
...
+			    </dependencyAnalysisParticipation>
+			</source>
+		    </sources>
+		</configuration>
+		<executions>
+		    <execution>
+			<id>javah</id>
+			<phase>generate-sources</phase>
+			<goals>
+			    <goal>javah</goal>
```
lukasj, line 161: this has to replaced by `javac -h` followed by `compile`

jbescos:
The goal javah can be skipped because the .h files are generated in mbox module. That solves the issue of building it with JDK 11.

I have updated the mbox/native/pom.xml file to use that .h files. And also I had to add a new property: linker.libs because that path is different in JDK 11.

I'm going to check that the generated libmbox.so is usable before I set it back to 'ready for review'

jbescos:
I have created one test that invokes a native method. I am not going to commit, because:
- It depends on the generated lib libmbox.so
- For example in Windows will not work

But I copy it here:

```
public class NativeTest {

    @Test
    public void test() {
        long value = UNIXFile.lastAccessed0("/home/jbescos/workspace/mail/mbox/native/target/libmbox-sources.jar");
        System.out.println("Value is: " + new Date(value));
    }
}

Running com.sun.mail.mbox.NativeTest
Value is: Fri Mar 26 15:17:50 CET 2021
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.036 sec
```

mbox/native/pom.xml
```diff
@@ -1,39 +1,165 @@
...
+		    ${env.MACH}
+		</linker.arch>
+		<linker.libs>${env.JAVA_HOME}/jre/lib/${linker.arch}</linker.libs>
+		<linker.end.options>
+		    -L${linker.libs} -lmail -ljava -lc
+		</linker.end.options>
+    </properties>
+
+    <profiles>
+	<profile>
```
pzygielo, line 64: With enforced rule
```
[WARNING] Rule 1: org.apache.maven.plugins.enforcer.RequireJavaVersion failed with message:
Detected JDK Version: 1.8.0-281 is not in the allowed range [11,).
```
perhaps this profile `1.8` could be removed?

jbescos: Yes, there is no point to have it because of the JDK version validator.

jbescos: Done, thanks for your finding

mbox/native/pom.xml
```diff
@@ -1,39 +1,148 @@
...
+    </properties>
+
+    <profiles>
+	   <profile>
+        <!--
+        Override the settings necessary to build with JDK 11.
+        -->
+        <id>11</id>
+        <activation>
+            <jdk>11</jdk>
```
lukasj, line 70: this profile is not needed since using JDK11 for build is required

jbescos: I have removed it, thank you.

lukasj (approved):

PR #541, opened 2021-03-22
Split mail in API and Impl

jbescos:
This PR is about being able to build API and Impl packages separately.

The most problematic part is that API was depending on some Impl classes. I had to bring that classes inside the API packages, so this will require a modification in the spec. These are the classes:
```
	renamed:    mail/src/main/java/com/sun/mail/util/ASCIIUtility.java -> mail/src/main/java/jakarta/mail/util/ASCIIUtility.java
	renamed:    mail/src/main/java/com/sun/mail/util/BASE64DecoderStream.java -> mail/src/main/java/jakarta/mail/util/BASE64DecoderStream.java
	renamed:    mail/src/main/java/com/sun/mail/util/BASE64EncoderStream.java -> mail/src/main/java/jakarta/mail/util/BASE64EncoderStream.java
	renamed:    mail/src/main/java/com/sun/mail/util/BEncoderStream.java -> mail/src/main/java/jakarta/mail/util/BEncoderStream.java
	renamed:    mail/src/main/java/com/sun/mail/util/DecodingException.java -> mail/src/main/java/jakarta/mail/util/DecodingException.java
	renamed:    mail/src/main/java/com/sun/mail/util/FolderClosedIOException.java -> mail/src/main/java/jakarta/mail/util/FolderClosedIOException.java
	renamed:    mail/src/main/java/com/sun/mail/util/LineInputStream.java -> mail/src/main/java/jakarta/mail/util/LineInputStream.java
	renamed:    mail/src/main/java/com/sun/mail/util/LineOutputStream.java -> mail/src/main/java/jakarta/mail/util/LineOutputStream.java
	renamed:    mail/src/main/java/com/sun/mail/util/LogOutputStream.java -> mail/src/main/java/jakarta/mail/util/LogOutputStream.java
	renamed:    mail/src/main/java/com/sun/mail/util/MailLogger.java -> mail/src/main/java/jakarta/mail/util/MailLogger.java
	renamed:    mail/src/main/java/com/sun/mail/util/MessageRemovedIOException.java -> mail/src/main/java/jakarta/mail/util/MessageRemovedIOException.java
	renamed:    mail/src/main/java/com/sun/mail/util/MimeUtil.java -> mail/src/main/java/jakarta/mail/util/MimeUtil.java
	renamed:    mail/src/main/java/com/sun/mail/util/PropUtil.java -> mail/src/main/java/jakarta/mail/util/PropUtil.java
	renamed:    mail/src/main/java/com/sun/mail/util/QDecoderStream.java -> mail/src/main/java/jakarta/mail/util/QDecoderStream.java
	renamed:    mail/src/main/java/com/sun/mail/util/QEncoderStream.java -> mail/src/main/java/jakarta/mail/util/QEncoderStream.java
	renamed:    mail/src/main/java/com/sun/mail/util/QPDecoderStream.java -> mail/src/main/java/jakarta/mail/util/QPDecoderStream.java
	renamed:    mail/src/main/java/com/sun/mail/util/QPEncoderStream.java -> mail/src/main/java/jakarta/mail/util/QPEncoderStream.java
	renamed:    mail/src/main/java/com/sun/mail/util/UUDecoderStream.java -> mail/src/main/java/jakarta/mail/util/UUDecoderStream.java
	renamed:    mail/src/main/java/com/sun/mail/util/UUEncoderStream.java -> mail/src/main/java/jakarta/mail/util/UUEncoderStream.java
```
Currently tests are passing, but I set this PR as a draft because I am going to try to make some sort of contract between the API and Impl with previous classes.


jbescos:
Split is done, but work is in progress because I need to add javadoc comments and make some improvements maven and module-info related.

You can start to take a look. Here is the summary of this change:
1. The impl can be found in mail/mail, and API in mail/mailapi
2. One of the problems is that there are many OutputStreams and InputStreams that were used by API and Impl, so I created a new package in API jakarta.mail.stream with interfaces that Impl is supossed to implement. The Impl must implement streams for every constant in StreamProvider.
3. I tried to avoid copy-paste but there are some places that I couldn't. However these duplicates in API are not public. Here is the list sorted aphabeticaly:
    ASCIIUtility#getBytes copy-pasted also in MimeUtilityUtility#getBytes
4. These are the classes that I was not able to keep in Impl, so they will be new classes in the API. It needs discussion about how to proceed with them:
    4.1 jakarta.mail.util.DefaultProvider -> This annotation is read by the Session in the API. The only way I see to get rid of this is to remove that functionality.
    4.2 jakarta.mail.util.FolderClosedIOException and MessageRemovedIOException -> These exceptions are a duplicate of FolderClosedException and MessageRemovedException, so it should be easy to remove. But it is not, because they are thrown in methods that have 'throws IOException' in the method signature. We can remove them if we make MessageException to extend IOException.
    4.3 jakarta.mail.util.PropUtil -> This is used to read properties. It makes sense to have it in API.
    4.4 jakarta.mail.util.MailLogger -> This one can be removed from the API. There are few classes that uses it, but it looks a nice to have feature and not relevant for the core functionality.

jbescos:
I think it is ready for other iteration of reviews.

lukasj (commented):
changes should be kept within boundaries implied by backwards compatibility requirements. Is there any technical reason for moving existing streams to the new package? If not, the new `StreamProvider` can be member of the `j.m.util` package.

wrt `StreamProvider` itself - what about having an enum instead of String constants? While it is true that adding new entry to enum is backward incompatible change, is it expected that something gets added there in forseeable future?

lukasj:
@jmehrens would be great if you could take a look at this while it's still in progress ;-)

jmehrens:
2. Are we creating new public types as a part of the API?  If so, for the stream classes could we make those types non-public and in the same package as j.i.MimeUtililty?  MimeUtility is the factory for most of these.  MimeUtil class code could be pushed into the MimeBodyPart class as package private static method.

4.1 - We might be able to just get rid of DefaultProvider.  I would think that Session.class.getClassloader() == XXXProvider.class.getClassloader() && XXXProvider.class.getName().startsWith("com.sun.mail.") would do the same thing as the annotation is trying to do.

4.2 jakarta.mail.util.FolderClosedIOException and MessageRemovedIOException could we just use exception chaining to work around including these types?  E.G new IOException(new FolderClosedException(...)) then walk the chain in getContent?

4.3. PropUtil  I think the session convivence methods could be pushed into the Session class and the System property methods should use the ones that ship with the JDK like [getInteger](https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#getInteger-java.lang.String-int-).  The only downside is there is no 2 argument Boolean.getBoolean.  We would have to use "true".equalsIgnoreCase(getProperty("key", "true")) for cases where the default is not specified equals true.  The only minor difference is that security exception is not handled but SecurityManager is going away in the JDK.

4.4 - The com.sun.mail.util.MailLogger was created when logging API support was added to JavaMail.  The purpose of this class is to make it so the Session.setDebugXXX methods function even if the JUL logger settings are silenced while still having the convenience of logging once at the callsite.  If we are targeting a min version of JDK9+ then we should ditch this type and update mail API to use [java.lang.System.LoggerFinder](https://docs.oracle.com/javase/9/docs/api/java/lang/System.LoggerFinder.html) and keep this bridge in the implementation.  If we are targeting min version of JDK8 then we should consider just using the java.util.logging.Logger over MailLogger.  The Session.setDebugXXX methods can then just install and remove a StreamHandler to act as the bridge.  It just means that logging can override the session debug methods.  I can build the patch for that off of this branch if needed to make it more clear.

jbescos:
@lukasj , @jmehrens , thank you for your suggestions.

2. The new public interfaces are supposed to be implemented by different implementations (then they need to be public). I moved to the package that @lukasj  said (jakarta.mail.util).

I implemented the points 4.1 and 4.2 in the way you said @jmehrens , it was a very good idea.

I had a review with @lukasj  some weeks ago, and we sorted out the points 4.3 and 4.4 already. It can be found in this commit:
https://github.com/eclipse-ee4j/mail/pull/541/commits/9eb08175a4b17efa773b128582298985a200b75a

Both classes (PropUtil and MailLogger) exists as public only in the implementation. Still, the API has a package private MailLogger.

mailapi/src/main/java/jakarta/mail/util/StreamProvider.java
```diff
@@ -0,0 +1,160 @@
...
+
+import java.io.InputStream;
+import java.io.OutputStream;
+
+/**
+ * Interface to be implemented by different mail implementations. It contains
+ * the methods to instance different encoders/decoders and other streams required by the API.
+ * 
+ * Implementations of this interface will need to be declared in
+ * META-INF/services/jakarta.mail.stream.StreamProvider
```
lukasj, line 27: this has to be made more general - sth like `Service lookup is used to find Implementations of this interface` (in the other words - current wording does not cover the case where the provider is defined in module-info)

mail/src/main/java/module-info.java
```diff
@@ -49,4 +45,7 @@
             com.sun.mail.smtp.SMTPSSLProvider,
             com.sun.mail.pop3.POP3Provider,
             com.sun.mail.pop3.POP3SSLProvider;
+    uses jakarta.mail.util.StreamProvider;
```
lukasj, line 48: this is needed when the module does the lookup for this service - I do not see it's the case here, so this line should be removed

lukasj (commented):
`SharedInputStream` have to stay in `j.m.internet` package

lukasj (approved):
LGTM

Issue #530, opened 2021-03-04
Address reliance on default constructors


jmehrens:
JDK16 issues a warning during complication when the compiler adds the default constructor.  For background see:
[JDK-8250212 - Address reliance on default constructors in the JDK (umbrella)](https://bugs.openjdk.java.net/browse/JDK-8250212)

For most cases a no argument public constructor should be added with javadoc.  

`jakarta/mail/Address.java:[32,17] class jakarta.mail.Address in exported package jakarta.mail declares no explicit constructors, thereby exposing a default constructor to clients of module jakarta.mail
jakarta/mail/Authenticator.java:[51,17] class jakarta.mail.Authenticator in exported package jakarta.mail declares no explicit constructors, thereby exposing a default constructor to clients of module jakarta.mail
jakarta/mail/search/SearchTerm.java:[41,17] class jakarta.mail.search.SearchTerm in exported package jakarta.mail.search declares no explicit constructors, thereby exposing a default constructor to clients of module jakarta.mail
jakarta/mail/BodyPart.java:[30,17] class jakarta.mail.BodyPart in exported package jakarta.mail declares no explicit constructors, thereby exposing a default constructor to clients of module jakarta.mail
jakarta/mail/event/ConnectionAdapter.java:[27,17] class jakarta.mail.event.ConnectionAdapter in exported package jakarta.mail.event declares no explicit constructors, thereby exposing a default constructor to clients of module jakarta.mail
jakarta/mail/search/ComparisonTerm.java:[26,17] class jakarta.mail.search.ComparisonTerm in exported package jakarta.mail.search declares no explicit constructors, thereby exposing a default constructor to clients of module jakarta.mail
com/sun/mail/handlers/handler_base.java:[25,17] class com.sun.mail.handlers.handler_base in exported package com.sun.mail.handlers declares no explicit constructors, thereby exposing a default constructor to clients of module jakarta.mail
com/sun/mail/handlers/text_plain.java:[28,8] class com.sun.mail.handlers.text_plain in exported package com.sun.mail.handlers declares no explicit constructors, thereby exposing a default constructor to clients of module jakarta.mail
jakarta/mail/event/TransportAdapter.java:[27,17] class jakarta.mail.event.TransportAdapter in exported package jakarta.mail.event declares no explicit constructors, thereby exposing a default constructor to clients of module jakarta.mail
[...]

PR #545, opened 2021-03-27
Address reliance on default constructors #530

jmehrens:
Fix for https://github.com/eclipse-ee4j/mail/issues/530

ComparisonTerm seem like it should have a private default or deprecated default and a new single argument constructor added.  For now I just left it as is for compatibility.  Seems like it could still use some more javadoc.

BASE64MailboxDecoder should have a private default constructor but for compatibility I included a default and deprecated it.

I added reflection tests to logging package to ensure the class and constructor modifiers are enforced.


lukasj (approved):
LGTM

PR #554, opened 2021-04-07
MethodHandles only supported in Android API version 26

jmehrens:
@lukasj Here is the patch for https://github.com/eclipse-ee4j/mail/issues/529 to fix the issue with MethodHandles on Android.  What test are you running to check for this error? Something on the build server?

### DIFF
--- a/mail/src/main/java/com/sun/mail/util/logging/LogManagerProperties.java
+++ b/mail/src/main/java/com/sun/mail/util/logging/LogManagerProperties.java
@@ -17,10 +17,6 @@
 package com.sun.mail.util.logging;
 
 import java.io.*;
-import java.lang.invoke.MethodHandle;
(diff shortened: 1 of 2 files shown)


lukasj (approved):
LGTM

> What test are you running to check for this error?

`mvn -U -C clean install -Poss-release -Dgpg.skip=true` is what I run locally and what is - w/o skipping gpg - also used to build snapshots/releases (it checks javadoc errors as well). Snapshot builds for [1x](https://ci.eclipse.org/mail/job/mail-v1x-snapshot/) as well as for [master](https://ci.eclipse.org/mail/job/mail-master-snapshot/) will be updated shortly to send heads up to dev-list. The failure in this particular case came from [sigtests](https://github.com/eclipse-ee4j/mail/blob/v1.x/android/pom.xml#L67)

PR #566, opened 2021-07-12
Add support for custom HostnameVerifier with Secure Sockets

icu5545:
This PR is to provide the ability to utilize a custom HostnameVerifier implementation when using a secure socket for mail transport. This provides the ability for consumers of the API to use a pre-existing hostname verification implementation as opposed to performing additional, potentially redundant, error-prone property configuration.

### DIFF
--- a/mail/src/main/java/com/sun/mail/util/SocketFetcher.java
+++ b/mail/src/main/java/com/sun/mail/util/SocketFetcher.java
@@ -607,7 +607,7 @@ public class SocketFetcher {
 	boolean idCheck = PropUtil.getBooleanProperty(props,
 			    prefix + ".ssl.checkserveridentity", false);
 	if (idCheck)
-	    checkServerIdentity(host, sslsocket);
+	    checkServerIdentity(props, prefix, host, sslsocket);
 	if (sf instanceof MailSSLSocketFactory) {
 	    MailSSLSocketFactory msf = (MailSSLSocketFactory)sf;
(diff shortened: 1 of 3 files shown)


jmehrens:
I think you should convert this code so that we are storing the string of the class name used to construct the HostnameVerifier.  I don't think we want to add a constructed HostnameVerifier to the properties object.  This would be similar to how we create services in the session class. https://github.com/eclipse-ee4j/mail/blob/master/mail/src/main/java/jakarta/mail/Session.java#L788  or how we create Folder subclasses in IMAP.

icu5545:
@jmehrens Agreed, a string-based option should definitely be made available. I'm basing the implementation on the existing code within the SocketFetcher. E.g. SocketFactory resolution: https://github.com/eclipse-ee4j/mail/blob/dc446219e3276878031cac02d6ff7b42a1e8717c/mail/src/main/java/com/sun/mail/util/SocketFetcher.java#L149-L183 

I'm not sure if you're suggesting it, but I'm not comfortable completely eliminating the ability to provide an object within the properties. First, the precedent is set, and keeps this implementation in line with existing functionality. And, second, because the HostnameVerifier interface does not provide any specifics into construction mechanics, imposing one-size-fits-all constructor or singleton handling does not seem to make sense.

jmehrens:
@icu5545 That seem fine to allow implementation and string based like socket factory.

jmehrens:
Moved the PR to https://github.com/eclipse-ee4j/angus-mail/pull/126

PR #575, opened 2021-08-19
#568: Move implementation out from the API repository

### LINKED ISSUE #568: Move implementation out from the API repository
To make sure that APIs are really independent, implementation needs to be moved to its own repository.

The new name for the implementation is going to be Eclipse Angus.

lukasj:
fixes #568 ; implementation has been moved to https://github.com/eclipse-ee4j/angus-mail repository


todorinskiz:
Shouldn't this be mentioned in the Changelog?

PR #578, opened 2021-08-27
Allow to override internal static fields

jbescos:
I have closed this PR https://github.com/eclipse-ee4j/mail/pull/515 and address it in this new PR. This is because of the conflicts after the split of API and Implementation.

### DIFF
--- a/api/src/main/java/jakarta/mail/internet/MimeMultipart.java
+++ b/api/src/main/java/jakarta/mail/internet/MimeMultipart.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2023 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -1003,8 +1003,8 @@ public class MimeMultipart extends Multipart {
      */
     protected MimeBodyPart createMimeBodyPart(InputStream is)
 				throws MessagingException {
-        if (ds != null && ds instanceof MimePartDataSource) {
-            Session session = ((MimePartDataSource)ds).getMessageContext().getSession();
+        if (ds instanceof MessageAware) {
+            Session session = ((MessageAware)ds).getMessageContext().getSession();
             if (session != null) {
                 return new MimeBodyPart(session, is);
             }


api/src/main/java/jakarta/mail/internet/MimePartDataSource.java
```diff
@@ -149,9 +159,7 @@ public String getName() {
      * @since JavaMail 1.1
      */
     @Override
-    public synchronized MessageContext getMessageContext() {
-	if (context == null)
-	    context = new MessageContext(part);
-	return context;
+    public MessageContext getMessageContext() {
```
lukasj, line 162: why was `synchronized` removed?

api/pom.xml
```diff
@@ -404,6 +404,8 @@ Use is subject to <a href="{@docRoot}/doc-files/speclicense.html" target="_top">
                 <configuration>
                     <forkCount>2</forkCount>
                     <reuseForks>false</reuseForks>
+                    <!-- Test service loader is not loaded if modules are enabled -->
+                    <useModulePath>false</useModulePath>
```
lukasj, line 408: for this to work, tests need their own module-info, there are more ways to make it work

lukasj (commented):
@jmehrens how does this look to you?

api/src/main/java/jakarta/mail/internet/MimePartDataSource.java
jbescos: There is no point to have it synchronized because the context is instanced now in the constructor. It is not possible that 2 threads obtain a different instance by calling that method.

api/pom.xml
jbescos: Added, thank you.

api/src/test/java/jakarta/mail/internet/MimeBodyPartTest.java
```diff
@@ -0,0 +1,88 @@
...
+        assertTrue(obp.allowutf8);
+        assertTrue(obp.cacheMultipart);
+        // Change the properties in opposite way
+        prop.put("mail.mime.setdefaulttextcharset", Boolean.FALSE);
+        prop.put("mail.mime.setcontenttypefilename", Boolean.FALSE);
+        prop.put("mail.mime.encodefilename", Boolean.TRUE);
+        prop.put("mail.mime.decodefilename", Boolean.TRUE);
+        prop.put("mail.mime.ignoremultipartencoding", Boolean.FALSE);
+        prop.put("mail.mime.allowutf8", Boolean.FALSE);
+        prop.put("mail.mime.cachemultipart", Boolean.FALSE);
```
jmehrens, line 65: If I recall correctly, we are expecting String key and String values.  It must work because toString gets called on the value.

api/src/main/java/jakarta/mail/internet/MimeBodyPart.java
```diff
@@ -216,6 +218,34 @@ public MimeBodyPart(InternetHeaders headers, byte[] content)
...
 
+    /**
+     * Initializes MimeBodyPart from the InputStream is and it overwrites
+     * properties from session.
+     *
+     * @param session   Session object for this message
+     * @param is    the message input stream
+     * @exception   MessagingException for failures
+     */
+    public MimeBodyPart(Session session, InputStream is) throws MessagingException {
```
jmehrens, line 229: Do we need a since tag?

jmehrens (approved):
I like the version.  Much cleaner than what we started with.

jbescos:
Notice that this changes breaks the TCK and I created a PR for that:
https://github.com/eclipse-ee4j/mail-tck/pull/24

It is about setting the system property at the begging of the test.

api/src/main/java/jakarta/mail/internet/MimeMultipart.java
```diff
@@ -1002,7 +1003,13 @@ protected MimeBodyPart createMimeBodyPart(InternetHeaders headers,
      */
     protected MimeBodyPart createMimeBodyPart(InputStream is)
 				throws MessagingException {
-	return new MimeBodyPart(is);
+        if (ds != null && ds instanceof MimePartDataSource) {
```
jmehrens, line 1006: I think this should simply be:
`if (ds instanceof MessageAware)`

followed by a cast to MessageAware instead of MimePartDataSource

jmehrens (changes requested):
For https://github.com/jakartaee/mail-api/issues/589 we'll need an override for InternetAddress utf8 boolean.

jbescos:
Setting this as a draft meanwhile this comment is not implemented: https://github.com/jakartaee/mail-api/pull/578#pullrequestreview-1231115312

PR #579, opened 2021-09-13
StreamProvider not static and OSGi support

jbescos:
StreamProvider is supposed to be implemented by different implementations, and it was instanced one time in a static variable.

This could be a problem in server containers, because only one instance would exist. It could happen that it is loaded as com.sun.mail, and later an application with other implementation is deployed.  In this case, this application would still use com.sun.mail.

### DIFF
--- a/api/src/main/java/jakarta/mail/util/FactoryFinder.java
+++ b/api/src/main/java/jakarta/mail/util/FactoryFinder.java
@@ -1,0 +1,151 @@
+/*
+ * Copyright (c) 2021 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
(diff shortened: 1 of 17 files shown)


api/src/main/java/jakarta/mail/util/FactoryFinder.java
```diff
@@ -0,0 +1,151 @@
...
+        throw new IllegalStateException("Not provider of " + factoryClass.getName() + " was found");
+    }
+
+    private static <T> T newInstance(String className) throws RuntimeException {
+        ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+        checkPackageAccess(className);
+        Class<T> clazz = null;
+        try {
+            if (classLoader == null) {
+                clazz = (Class<T>) Class.forName(className);
```
lukasj, line 73: unchecked cast... add appropriate suppresswarnings....

lukasj (commented):
with activation-api 2.1.0-RC2 needs angus-activation at test scope (or some dummy impl probably...) to build

lukasj (approved):
I'll fix outstanding minor things myself in the next PR

basil:
Causes downstream problems for Jenkins: #665

PR #580, opened 2021-09-15
End MIME boundary should be followed by CRLF

joshihirenn:
The present code mandates that a CRLF is present after intermediate MIME boundary (that doesn't end with --). This change extends that behaviour for end boundary too (that starts and ends with --).
This is useful for processing some bulk emails that uses a single dash character as a boundary and a line full of dashes is present in the content. Gmail and Outlook are able to process such emails, however JavaMail truncates the email after the sequence of five dash characters are consumed from the line full of dash characters.

Signed-off-by : The author is aware of the terms by which the contribution has been provided to the project.

### DIFF
--- a/api/src/main/java/jakarta/mail/internet/MimeMultipart.java
+++ b/api/src/main/java/jakarta/mail/internet/MimeMultipart.java
@@ -806,9 +806,25 @@ public class MimeMultipart extends Multipart {
 			    int b2 = in.read();
 			    if (b2 == '-') {
 				if (in.read() == '-') {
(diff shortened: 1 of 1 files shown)


lukasj (commented):
Can you add a test (simple junit is fine) for this as well, please?

joshihirenn:
> Can you add a test (simple junit is fine) for this as well, please?

joshihirenn:
> > Can you add a test (simple junit is fine) for this as well, please?

Sure, I will try to add one.

PR #597, opened 2022-03-15
Implement equals() and hashcode() on jakarta.mail.Header

jbescos:
https://github.com/eclipse-ee4j/mail/issues/594

### DIFF
--- a/api/src/main/java/jakarta/mail/Header.java
+++ b/api/src/main/java/jakarta/mail/Header.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2022 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -16,6 +16,7 @@
 
 package jakarta.mail;
 
+import java.util.Objects;
 
 /**
  * The Header class stores a name/value pair to represent headers.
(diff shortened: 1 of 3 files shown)


jbescos:
It fails because of copyright in unrelated files. Strangely, it does not fail when I run the same command locally:
`mvn -B -U -C -V clean verify org.glassfish.copyright:glassfish-copyright-maven-plugin:check -Poss-release,staging -Dgpg.skip=true `

lukasj:
add `-Dcopyright.ignoreyear=true` to the build command in the GH action

api/src/main/java/jakarta/mail/Header.java
```diff
@@ -67,4 +67,36 @@ public String getName() {
     public String getValue() {
 	return value;
     }
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
```
pzygielo, line 73: Could `java.util.Objects.hash(name, value)` be used instead?

api/src/main/java/jakarta/mail/Header.java
```diff
@@ -67,4 +67,36 @@ public String getName() {
...
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj)
+            return true;
+        if (obj == null)
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
+        Header other = (Header) obj;
+        if (name == null) {
```
pzygielo, line 89: Could `java.util.Objects.equals(name, other.name) && java.util.Objects.equals(value, other.value)` be used instead?

jbescos: For the equals I have to verify first that `obj` is not null and it is the same class before using `Objects.equals`.

api/src/main/java/jakarta/mail/Header.java
jbescos: Sure

jbescos:
Same error than here:

https://github.com/eclipse-ee4j/mail/pull/599#issuecomment-1087213759

lukasj (commented):
put a note to changes, please

jbescos:
> put a note to changes, please

Done

lukasj (approved):
LGTM

Issue #621, opened 2022-09-27
j.m.u.FactoryFinder.factoryFromServiceLoader needs PrivilegedAction


boris-unckel:
**Describe the bug**
Using the service load approach jakarta.mail.util.FactoryFinder.factoryFromServiceLoader with an active security fails in an AppServer environment. The user space app would be forced to grant the permission. This is what we got in WildFly  27:

Here is the stack trace https://gist.github.com/jamezp/0c413ca12a51c5f39843702cdd003ddb

This is where the issue would be https://github.com/eclipse-ee4j/mail/blob/c424303d5dafab2921c3893085a6662f69550ef5/api/src/main/java/jakarta/mail/util/StreamProvider.java#L176-L178. Specifically this https://github.com/eclipse-ee4j/mail/blob/c424303d5dafab2921c3893085a6662f69550ef5/api/src/main/java/jakarta/mail/util/FactoryFinder.java#L130. That would need to be in a privileged action.

**Expected behavior**
Use a PrivilegedAction to load the service, sample code here: https://github.com/eclipse-ee4j/mail/blob/c424303d5dafab2921c3893085a6662f69550ef5/api/src/main/java/jakarta/mail/util/FactoryFinder.java#L90 ) 

**Additional context**
A cleaner approach could be a spec defined permission check on accessing the mail API i.e. a general permission we could apply to the deployment - then internally once that permission check has been performed use a doPrivileged to drop the callers protection domain from the specific permissions check as the caller does not need to know about the details.

Discussion about the bug started here
https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/Jakarta.20Mail

scottmarlow:
If someone creates a pull request that addresses this issue and the change is approved, could that change be included in a Jakarta Mail 2.1.1 service release?

PR #623, opened 2022-10-04
Add the ability to the API consumers to load the API implementations …

yersan:
…by using a different protection domain when the API is used with the security manager enabled

Complete the fix for #621

#621 was fixed by adding a new privilege block for the point where the iterator.next was being called, however, looking more closely at the different methods under FactoryFinder#find, there are other points in the path that can also fire SM checks, for example `Thread.currentThread().getContextClassLoader();` or `Class.forName(OSGI_SERVICE_LOADER_CLASS_NAME);`

This PR moves the whole doPrivilegde one level above the `StreamProvider` class which is indeed the one that explicitly loads the expected interface provided by the implementations.

The PRs roll back the changes added in https://github.com/eclipse-ee4j/mail/pull/622, the privilege block for the system property, and adds the check one level above.


jbescos:
Could you also add a new line here please?:
https://github.com/yersan/javamail/blob/1d500391b8e5358b7b7330d385f30cea03eccf08/doc/release/CHANGES.txt#L22

It should contain the next:
`E  623	Add the ability to the API consumers to load the API implementations`

The PR looks good to me.

yersan:
Hi @jbescos ,  I've just added the line to CHANGES.txt, thanks for the review.

jbescos:
Thank you @yersan , there is only other comment about the copyright year in the code review. Could you check it so I can approve this, please?.

yersan:
@jbescos ahh, right, for the StreamProvider.java. Sorry, I forgot to change that one. In any case, I didn't see such a comment in the code review though. It should be fine now, thanks.

jbescos (approved):
Thank you

lukasj (changes requested):
use spaces instead of TABs, please

yersan:
Hi @lukasj , done, notice the file in question has a mix of spaces vs tabs, I wasn't sure about the policy here. In general, the project has a mix of them in several places, a bulk update on all of them could be useful for the future, of course in a different issue

lukasj:
I agree usage of spaces/TABs is currently inconsistent in this project and we're trying to move to spaces to be consistent with most projects as we modify the code. If you want to bulk update it, feel free to prepare the PR for that. Thanks.

yersan:
Thanks. @lukasj, do you know by any chance when will be the next release of these two Mail and Activation?

Issue #630, opened 2022-10-21
Use OSGi service loader mediator instead of osgi resource locator


mnlipp:
To separate the API and the implementation, the service loader pattern is used. This perfectly well suited mechanism is known to have problems in OSGi environments.

An attempt has been made in the jakarta mail api implementation to solve the problems with org.glassfish.hk2.osgi-resource-locator, which requires lots of special coding.

The "service loader and OSGi" problem has actually successfully been addressed before by http://docs.osgi.org/specification/osgi.enterprise/7.0.0/service.loader.html with the reference implementation (with an easier to read description) [SPI Fly](https://aries.apache.org/documentation/modules/spi-fly.html). What it boils down to is that instead of having to maintain the complicated code, you can simply use the service locator after adding some statements to MANIFEST.MF.

While this is not a runtime bug, because you can use Jakarta mail api in an OSGi environment (once you have found out that you need to include the hk2 jar and once https://github.com/eclipse-ee4j/mail/issues/629 will be fixed), I think it is a bug with respect to the design.

mnlipp:
Having found the usage of hk2 to be implemented [only partially](https://github.com/eclipse-ee4j/mail/issues/629#issuecomment-1287850248), I have patched the `MANIFEST.MF` files of the jars as required for the use of the service loader mediator. I did it manually, as I don't know which build system is used. The required headers are:

In `jakarta.mail-api`:
```properties
Require-Capability: osgi.extender;filter:="(osgi.extender=osgi.servicelo
 ader.processor)";resolution:=optional,osgi.ee;filter:="(&(osgi.ee=JavaS
 E)(version=1.8))"
```

In `angus-core`:
```properties
Provide-Capability: osgi.serviceloader;osgi.serviceloader="jakarta.mail.
 util.StreamProvider"
Require-Capability: osgi.extender;filter:="(osgi.extender=osgi.servicelo
 ader.registrar)",osgi.extender;filter:="(osgi.extender=osgi.serviceload
 er.processor)";resolution:=optional,osgi.ee;filter:="(&(osgi.ee=JavaSE)
 (version=1.8))"
```

And in `imap` (using this as an example):
```properties
Provide-Capability: osgi.serviceloader;osgi.serviceloader="jakarta.mail.
 Provider"
Require-Capability: osgi.extender;filter:="(osgi.extender=osgi.servicelo
 ader.registrar)",osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
```

[...]

PR #632, opened 2022-10-23
Using service loader mediator annotations (see #630).

mnlipp:
Shows the usage of the [service locator mediator](https://github.com/eclipse-ee4j/mail/issues/630) with annotations that generate the "Require-Capability" header in MANIFEST.MF. (Alternatively the Header can be specified directly in pom.xml.)

The Providers (in angus-mail) need a corresponding `@ServiceProvider` annotation such as:
```
@aQute.bnd.annotation.spi.ServiceProvider(value = StreamProvider.class)
public class MailStreamProvider implements StreamProvider {
```

(Sorry, can't provide a PR for that, maven compilation for angus-mail doesn't work.)

### DIFF
--- a/api/src/main/java/jakarta/mail/Session.java
+++ b/api/src/main/java/jakarta/mail/Session.java
@@ -198,6 +198,7 @@ import jakarta.mail.util.StreamProvider;
  * @author Max Spivak
  */
 
+@aQute.bnd.annotation.spi.ServiceConsumer(value = Provider.class)
 public final class Session {
 
 	// Support legacy @DefaultProvider
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -116,6 +116,18 @@
             <artifactId>angus-activation</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>biz.aQute.bnd</groupId>
+            <artifactId>biz.aQute.bnd.annotation</artifactId>
+            <version>6.3.1</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.osgi</groupId>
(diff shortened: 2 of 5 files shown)


jbescos:
I wrote in the issue by mistake. Find the comment here:
https://github.com/jakartaee/mail-api/issues/630#issuecomment-1290056613

mnlipp:
> Ideally there should be a plugin to deal with the manifest. 

You don't need an extra plugin. You can simply add this information using the Felix maven bundle plugin which is already in pom.xml with instructions `<Provide-Capability>` and `<Require-Capability>` (that's doing it "manually", see https://github.com/jakartaee/mail-api/issues/630#issuecomment-1287852918). In general, the manual approach is considered a bit more difficult to maintain, but once it's there, it rarely changes (only if you add another loaded service).

If it helps, I can provide another PR with the pom.xml based solution.

api/src/main/java/jakarta/mail/Session.java
```diff
@@ -198,6 +198,7 @@
  * @author Max Spivak
  */
 
+@aQute.bnd.annotation.spi.ServiceConsumer(value = Provider.class)
```
lukasj, line 201: There is currently a restriction on using 3rd party libraries in the APIs and only (non Java SE/Jakarta EE packages can be used.

lukasj (changes requested):
It is currently not allowed to depend on 3rd party libraries from APIs

mnlipp:
> It is currently not allowed to depend on 3rd party libraries from APIs

This is why this has actually been superseded by https://github.com/jakartaee/mail-api/pull/633.

lukasj:
so this should have been closed then

Issue #629, opened 2022-10-21
jakarta.mail-api-2.1.0.jar does not work in OSGi environment


mnlipp:
**Describe the bug**
An attempt has been made to solve service loader problems in an OSGi environment with org.glassfish.hk2.osgi-resource-locator.

The presence of the required class `org.glassfish.hk2.osgiresourcelocator.ServiceLoader` is detected with the code
```java
    private static boolean isOsgi() {
        try {
            Class.forName(OSGI_SERVICE_LOADER_CLASS_NAME);
            return true;
        } catch (ClassNotFoundException ignored) {
        }
        return false;
    }
```

In order for this to work, class `org.glassfish.hk2.osgiresourcelocator.ServiceLoader` must be in the bundle's classpath (jar becomes "bundle" in OSGi). But the required import is missng in `MANIFEST.MF` and therefore the provider services are not found.

**How it should be**
For comparison: `jakarta.activation-api-2.1.0.jar` has in its `MANIFEST.MF` the line:
```
DynamicImport-Package: org.glassfish.hk2.osgiresourcelocator
```
which ensures that the class `org.glassfish.hk2.osgiresourcelocator.ServiceLoader` is in the classpath and the `Class.forName` is successful.

**To Reproduce**
1. Deploy jakarta.mail-api-2.1.0.jar in an OSGi enviroment
2. Attempt to use the IMAP provider

**Expected behavior**
Provider should be found.

mnlipp:
Digging further, I found that the hk2 service locator is used in [FactoryFinder](https://github.com/eclipse-ee4j/mail/blob/82dae3a682f8b8a75dc426c28c458ffbcc78014f/api/src/main/java/jakarta/mail/util/FactoryFinder.java#L89) only. `Session.java` uses the `ServiceLoader` [without the hk2 "warpper"](https://github.com/eclipse-ee4j/mail/blob/82dae3a682f8b8a75dc426c28c458ffbcc78014f/api/src/main/java/jakarta/mail/Session.java#L984) and therefore fails to load the providers.

jbescos:
Hello,

It was reported one OSGi issue here:
https://github.com/eclipse-ee4j/angus-mail/issues/15

There is already a fix here, but not merged yet:
https://github.com/eclipse-ee4j/angus-mail/pull/19

Do you think that fix is valid here?.

mnlipp:
> Do you think that fix is valid here?.

[...]

PR #633, opened 2022-10-25
Fix630 2

mnlipp:
Implements https://github.com/jakartaee/mail-api/issues/630 (API side).

A PR without using annotations. It removes the hk2 code and adds to the pom.xml only.

Things look a bit complicated because the service mediator uses the general OSGi require/provide capability mechanism (not something "service loader" specific) for its configuration. Here's an explanation of the Require-Capability instruction added:
```
osgi.extender;filter:="(&(osgi.extender=osgi.serviceloader.processor)
                            (version>=1.0.0)(!(version>=2.0.0)))"
```
Tells the OSGi runtime that "something" must provide (in the namespace "osgi.extender") the capability "osgi.serviceloader.processor" with a version 1.x.x. This is the most general form of requiring something in OSGi. Usually, you require bundles with specific symbolic names (mostly through imports) because bundles provide an API or implementation that you call in your program and the symbolic name reflects the package name. But in this case we require a "feature" to be added to the OSGi runtime that we don't invoke (it has no API). It just should be there and do its job in the background. The feature in question is "properly handling service loader usage in OSGi" ("osgi.serviceloader.processor" for short) which will at runtime be provided to the OSGi runtime environment e.g. by adding the [SPI Fly](https://aries.apache.org/documentation/modules/spi-fly.html) jars.

The service loader (mediator) is available now, so let's configure it:
```
osgi.serviceloader;
    filter:="(osgi.serviceloader=jakarta.mail.util.StreamProvider)";cardinality:=multiple,

```
We tell the service loader (addressing it by using the namespace "osgi.serviceloader") that it should intercept all calls of `ServiceLoader.load` with `jakarta.mail.util.StreamProvider.class` as argument and do its magic (i.e. return the implementing classes despite the usual classloader isolation in OSGi). Note that special handling of the service loader is not specific to OSGi. You need a similar mechanism in [other module systems](https://blog.frankel.ch/migrating-serviceloader-java-9-module-system/) as well. `cardinality:=multiple' states that there may be several service implementations and the application code must choose (or use them all).

Finally we have:
```
osgi.serviceloader;
    filter:="(osgi.serviceloader=jakarta.mail.Provider)";cardinality:=multiple
```
because the service loader is used for two interfaces (two kinds of service) in jakarta-mail-api.

In order for this to work, the providers of services have to add some information to their `MANIFEST.MF` as well. I cannot get angus-mail to build (`maven package` results in an error and I know almost nothing about maven). So here's what must be added to the manifests of the provider bundles by adding instructions to pom.xml as above (tested by patching `MANIFEST.MF` in the angus-mail jar).

First, we (again) need a `<Require-Capability>`:
```
osgi.extender;filter:="(&(osgi.extender=osgi.serviceloader.registrar)(version>=1.0)(!(version>=2.0)))"
```
This makes sure that the bundle is processed by the platform "addition" `osgi.serviceloader.registrar` (provided by the SPI Fly jars). The registrar looks for service implementations and makes them available to the (intercepted) `ServiceLoader.load` invocations. We also have to specify which interfaces or classes actually define services to be consumed by the service loader, so we need a `<Provide-Capability>` instruction with:
```
osgi.serviceloader;osgi.serviceloader="jakarta.mail.util.StreamProvider";uses:="jakarta.mail.util",
osgi.serviceloader;osgi.serviceloader="jakarta.mail.Provider";uses:="jakarta.mail"
```
There's one entry for each file in `META-INF/services`. However, while you must explicitly specify the providing classes in the files in `META-INF/service/...`, the registrar finds them automatically while processing the bundle.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -364,10 +364,10 @@
                             (version>=1.0.0)(!(version>=2.0.0)))",
                           osgi.serviceloader;
                             filter:="(osgi.serviceloader=jakarta.mail.util.StreamProvider)";
-                              cardinality:=multiple;resolution:=optional,
+                              cardinality:=multiple,
                           osgi.serviceloader;
                             filter:="(osgi.serviceloader=jakarta.mail.Provider)";
-                              cardinality:=multiple;resolution:=optional]]>
+                              cardinality:=multiple]]>
                         </Require-Capability>
                     </instructions>
                 </configuration>
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -116,6 +116,18 @@
             <artifactId>angus-activation</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>biz.aQute.bnd</groupId>
+            <artifactId>biz.aQute.bnd.annotation</artifactId>
+            <version>6.3.1</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.osgi</groupId>
+            <artifactId>osgi.annotation</artifactId>
+            <version>8.1.0</version>
+            <scope>provided</scope>
+        </dependency>
     </dependencies>
 
     <build>
@@ -359,10 +371,6 @@
                         <Implementation-Title>${project.name}</Implementation-Title>
                         <Implementation-Vendor>${project.organization.name}</Implementation-Vendor>
                         <Implementation-Build-Id>${buildNumber}</Implementation-Build-Id>
-                        <Import-Package>
-                            !org.glassfish.hk2.osgiresourcelocator,
-                            *
-                        </Import-Package>
                     </instructions>
                 </configuration>
                 <executions>
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -359,6 +359,10 @@
                         <Implementation-Title>${project.name}</Implementation-Title>
                         <Implementation-Vendor>${project.organization.name}</Implementation-Vendor>
                         <Implementation-Build-Id>${buildNumber}</Implementation-Build-Id>
+                        <Import-Package>
+                            !org.glassfish.hk2.osgiresourcelocator,
+                            *
+                        </Import-Package>
                         <Require-Capability><![CDATA[
                           osgi.extender;filter:="(&(osgi.extender=osgi.serviceloader.processor)
                             (version>=1.0.0)(!(version>=2.0.0)))",
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -116,18 +116,6 @@
             <artifactId>angus-activation</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>biz.aQute.bnd</groupId>
-            <artifactId>biz.aQute.bnd.annotation</artifactId>
-            <version>6.3.1</version>
-            <scope>provided</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.osgi</groupId>
-            <artifactId>osgi.annotation</artifactId>
-            <version>8.1.0</version>
-            <scope>provided</scope>
-        </dependency>
     </dependencies>
 
     <build>
@@ -371,6 +359,16 @@
                         <Implementation-Title>${project.name}</Implementation-Title>
                         <Implementation-Vendor>${project.organization.name}</Implementation-Vendor>
                         <Implementation-Build-Id>${buildNumber}</Implementation-Build-Id>
+                        <Require-Capability><![CDATA[
+                          osgi.extender;filter:="(&(osgi.extender=osgi.serviceloader.processor)
+                            (version>=1.0.0)(!(version>=2.0.0)))",
+                          osgi.serviceloader;
+                            filter:="(osgi.serviceloader=jakarta.mail.util.StreamProvider)";
+                              cardinality:=multiple;resolution:=optional,
+                          osgi.serviceloader;
+                            filter:="(osgi.serviceloader=jakarta.mail.Provider)";
+                              cardinality:=multiple;resolution:=optional]]>
+                        </Require-Capability>
                     </instructions>
                 </configuration>
                 <executions>
--- a/api/src/main/java/jakarta/mail/Session.java
+++ b/api/src/main/java/jakarta/mail/Session.java
@@ -198,6 +198,7 @@ import jakarta.mail.util.StreamProvider;
  * @author Max Spivak
  */
 
+@aQute.bnd.annotation.spi.ServiceConsumer(value = Provider.class)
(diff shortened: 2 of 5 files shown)


jbescos (approved):

lukasj (changes requested):
SPI Fly is not available in GlassFish (yet?), so the API jar will stop working there after this change. Is there a way for both approaches to co-exist? Alternatively what changes need to be done on the GlassFish side to support this approach, preferably with keeping backward compatibility?

mnlipp:
> SPI Fly is not available in GlassFish (yet?), so the API jar will stop working there after this change. Is there a way for both approaches to co-exist? Alternatively what changes need to be done on the GlassFish side to support this approach, preferably with keeping backward compatibility?

I don't know enough about GlassFish. I always thought it is a "J2EE server" (old name, sorry, I stopped using J2EE a while ago).

The change does not require SPI Fly to be available in a non-OSGi environment.The is no dependency whatsoever. It' just a few declarations that take effect in OSGi environments.

Is GlassFish itself built on OSGi?

lukasj:
> Is GlassFish itself built on OSGi?

Yes, it is (it used to support felix, equinox and knopplerfish runtimes) and that is the main reason behind HK2 resource loader support in many APIs as a way to support service loader look up.

mnlipp:
> Is GlassFish itself built on OSGi?

Okay, I just found that GlassFish "includes" Apache Felix (quite a lot of 3rd party libraries...)

So if you use GlassFisch (actually Felix) as your OSGi environment then currently, you have to add the hk2-bundle to your runtime configuration. Else it won't work (it actually doesn't even work then because of https://github.com/jakartaee/mail-api/issues/629#issuecomment-1287850248 -- this will have to be fixed if this PR isn't accepted).

After merging this pull request, all you have to do is add the SPI Fly Bundle(s) to your list of runtime bundles. I suppose GlassFish doesn't use only the Apache Felix OSGi "core" framework, it also uses additional bundles from the "Apache Felix Ecosystem" that provide features such as SCR or the metadata service, config admin etc. The only difference between adding these bundles at runtime and adding the SPI Fly bundles is that the SPI Fly bundles aren't provided by "Apache  Felix" but by by "Apache Aries". Don't ask me why this particular aspect of implementing OSGi specifications has it's own name instead of being an Apache Felix subproject.

lukasj:
> > Is GlassFish itself built on OSGi?
> 
> Okay, I just found that GlassFish "includes" Apache Felix (quite a lot of 3rd party libraries...)
> 
> So if you use GlassFisch (actually Felix) as your OSGi environment then currently, you have to add the hk2-bundle to your runtime configuration. 

HK2 is there "by default", it is one of the core libraries around which GF was originally built (in v3, I think)

> 
> After merging this pull request, all you have to do is add the SPI Fly Bundle(s) to your list of runtime bundles. I suppose GlassFish doesn't use only the Apache Felix OSGi "core" framework, it also uses additional bundles from the "Apache Felix Ecosystem" that provide features such as SCR or the metadata service, config admin etc.

see [the core pom](https://github.com/eclipse-ee4j/glassfish/blob/master/nucleus/parent/pom.xml#L472-L566) for what is there and what is not


> The only difference between adding these bundles at runtime and adding the SPI Fly bundles is that the SPI Fly bundles aren't provided by "Apache Felix" but by by "Apache Aries". Don't ask me why this particular aspect of implementing OSGi specifications has it's own name instead of being an Apache Felix subproject.

GF would need to support (=include) Apache Aries "by default" first to allow APIs to remove integration of the HK2 resource loader

mnlipp:
> GF would need to support (=include) Apache Aries "by default" first to allow APIs to remove integration with the HK2 resource loader

Really, I don't care about politics. I think using the "standard" OSGi solution is the better way to go. But the only point that I'm really interested in is that the mail API continues to work in OSGi environments. So if you close this PR, please consider giving appropriate priority to the problem mentioned in https://github.com/jakartaee/mail-api/issues/629#issuecomment-1287850248. This PR would have solved the problem mentioned there as a side effect. But as this PR does not seem to happen, I've added an issue for the problem mentioned in the comment so that it get the necessary attention (https://github.com/jakartaee/mail-api/issues/635).

lukasj:
> I think using the "standard" OSGi solution is the better way to go.

I'm not saying that I disagree with the support for the better way. All I'm saying is - can both co-exist together? If so, why not just add what is required while keeping the rest untouched?

mnlipp:
> > I think using the "standard" OSGi solution is the better way to go.
> 
> I'm not saying that I disagree with the support for the better way. All I'm saying is - can both co-exist together? If so, why not just add what is required while keeping the rest untouched?

They can co-exist if you do a "find first (any)" search for services and if you keep strictly to the usage pattern as found in `FactoryFinder`. First do the "normal" `ServiceLoader` invocation (https://github.com/jakartaee/mail-api/blob/82dae3a682f8b8a75dc426c28c458ffbcc78014f/api/src/main/java/jakarta/mail/util/FactoryFinder.java#L52-L54). This will succeed if (a) you are not in an OSGi environment or (b) you are in an OSGI environment with a service loader mediator (and this PR has been applied). Else, the hk2 service locator will "kick in" (https://github.com/jakartaee/mail-api/blob/82dae3a682f8b8a75dc426c28c458ffbcc78014f/api/src/main/java/jakarta/mail/util/FactoryFinder.java#L58) (btw this should really be `isHk2Available()`, not `isOsgi()`).

Keep in mind, however, that using the service loader mediator must be actively supported by the provider of the service as well, because it must add the `Require-Capability`/`Provide-Capability` lines in its `MANIFEST.MF` as outlined at the beginning. So adding the service mediator support in the API will only have an effect if this is also supported by the developers of angus-mail (or alternatives). (But I assume I'm talking about the same people here.)

lukasj:
not only angus-mail, but also activation/angus-activation at least (since mail depends on that). Back to this particular PR - to accept it, update it as follows, please:

* keep addition of Require-Capability in pom
* rename `isOsgi` to `isHk2Available` - optional
* remove all other changes

so if service loader mediator is not supported, hk2 resource locator is kicked in as a fallback

thanks!

mnlipp:
Changed as requested.

lukasj:
Thank you! The only missing part now is ECA which needs to be signed in order to accept this, see [CONTRIBUTING.md](https://github.com/jakartaee/mail-api/blob/master/CONTRIBUTING.md#eclipse-contributor-agreement) for details.

mnlipp:
> Thank you! The only missing part now is ECA which needs to be signed in order to accept this, see [CONTRIBUTING.md](https://github.com/jakartaee/mail-api/blob/master/CONTRIBUTING.md#eclipse-contributor-agreement) for details.

Done.

PR #634, opened 2022-11-01
Fix wording in README.md

maeichris:
Minor fix to wording regarding the direct successor to JavaMail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,12 +6,12 @@
 Jakarta Mail defines a platform-independent and protocol-independent
 framework to build mail and messaging applications.
 
-**IMPORTANT:** Implementation of the Jakarta Mail API, aka JakartaMail (formerly JavaMail),
+**IMPORTANT:** The implementation of the Jakarta Mail API, aka JakartaMail (formerly JavaMail),
 is no longer part of this repository.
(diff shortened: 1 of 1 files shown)


lukasj (changes requested):
ECA needs to be signed in order to accept this, see [CONTRIBUTING.md](https://github.com/jakartaee/mail-api/blob/master/CONTRIBUTING.md#eclipse-contributor-agreement) for details.

maeichris:
ECA has been accepted.  "You successfully submitted an Eclipse Contributor Agreement."  Back here I clicked on "details" and told it to revalidate, which seems to have been successful so it appears to no longer have the warning.

Reading [the handbook](https://www.eclipse.org/projects/handbook/#resources-commit), it appears there are fields to associate with a commit; I am not sure where I should have specified them, or if any of them were supposed to be attached automatically.  Let me know how I can fix what I have done wrong, or if it is okay for this case but should be done differently in the future.  Sorry for the trouble.

lukasj (approved):
everything is fine now. I think the validation action is able to grab the email to check from the commit, so there should be no need to add sign-off (or any other) header to the commit message (required step in the past)

PR #638, opened 2022-11-04
Build last mail-tck bundle

jbescos:
Jenkins does not provide the bundled mail-tck, so in this PR it will build it from master.


lukasj (approved):
LGTM

PR #655, opened 2023-01-21
fix: factoryFromServiceLoader

cssxsh:
If mail-api and implement is loaded from the URLClassloader as a plugin, implement will not be found from ContextClassLoader


cssxsh:
May be related to # 550

Issue #660, opened 2023-02-06
jakarta.mail-api is not a Multi Release JAR


aalmiray:
**Describe the bug**
Latest jakarta.mail-api (`2.1.1`) includes 6 classes compiled to bytecode 53 and 148 classes to compiled to bytecode 52. All classes are found in the unversioned space which means enforcer rules for both [Maven](https://www.mojohaus.org/extra-enforcer-rules/enforceBytecodeVersion.html) and [Gradle](https://kordamp.org/enforcer-gradle-plugin/#_enforcebytecodeversion) will trigger and fail a build.

**To Reproduce**
1. Get [jarviz](https://github.com/kordamp/jarviz/releases)
```
$ jarviz bytecode show --gav jakarta.mail:jakarta.mail-api:2.1.1 
Unversioned classes. Bytecode version: 52 total: 148
Unversioned classes. Bytecode version: 53 total: 6
```

If the `--details` flag is added you get the names of the 6 classes causing trouble

```
Unversioned classes. Bytecode version: 53 total: 6
jakarta.mail.search.package-info
jakarta.mail.util.package-info
jakarta.mail.event.package-info
jakarta.mail.internet.package-info
jakarta.mail.package-info
module-info
```

**Expected behavior**
These 6 classes should be inside `/META-INF/versions/9`. The JAR's manifest should have a manifest entry such as `Multi-Release: true`

https://docs.oracle.com/javase/10/docs/specs/jar/jar.html#multi-release-jar-files

jbescos:
It looks we have the same issue we had here:
https://github.com/eclipse-ee4j/parsson/issues/73

I will apply the same fix.

FredDietrich:
What about module-info.class? It can neither be compiled nor ran with jdk/runtime 8. So the problem is only partially fixed with the package-info job?

Also, would (and could) this be ported over to 1.x? Asking because in need of multiline POP3 xoauth2 authentication, but, stuck to pre-jakarta mail (1.6.2) because of the classes compiled with 53 major version.

jbescos:
This can be addressed [here](https://github.com/jakartaee/mail-api/issues/709).

It looks we will not compile with JDK 8 anymore, everything will be compiled for JDK 9 or 11, and we don't need a multirelease for module-info.

jbescos:
[...]

PR #661, opened 2023-02-08
jakarta.mail-api is not a Multi Release JAR #660

jbescos:
https://github.com/jakartaee/mail-api/issues/660

```
$ mvn clean install && javap -verbose ./target/classes/jakarta/mail/internet/package-info.class | grep 'major version'
...
  major version: 52
```


api/pom.xml
```diff
@@ -337,6 +338,7 @@
                         </goals>
                         <configuration>
                             <release>8</release>
+                            <createMissingPackageInfoClass>true</createMissingPackageInfoClass>
```
lukasj, line 341: we don't want to create "empty" files

jbescos: If I set it to false, it does not generate any package-info.class at all.

It should be fine to keep it as it is, as long as we don't have packages with no package-info.java

However this seems to be a bug with the maven compiler plugin, and this is a workaround to bypass that.

jbescos: It is set to false now, because as you told me, it does not create the .class but it generates the javadoc. And in case something is added in the package-info.java, it generates the package-info.class

lukasj (approved):
LGTM

Issue #687, opened 2023-05-10
ParameterList fails to parse filename from Content-Disposition header encoded in UTF-8 with Q encoding


Dr4K4n:
**Describe the bug**
We are using the mail-api to parse incoming emails (MimeMessages), we received a particular email with a PDF attachment.
The filename of this attachment is encoded in the Content-Disposition header in a "weird" way.
This leads to the following exception

Stacktrace:
```
 Caused by: jakarta.mail.internet.ParseException: In parameter list <;
   filename==?utf-8?Q?XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX--111111111-XXXXXXXXXXXXXXXXXXX?=
   =?utf-8?Q?XXXXXXXXXXXXXXXXXXX=2Epdf?=;
   filename*0*=utf-8''XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX--111111111-XXXXXXXXXXX;
   filename*1*=XXXXXXXXXXXXXXXXXXXXXXXXXXX.pdf>, expected parameter value, got "="
     at jakarta.mail.internet.ParameterList.<init>(ParameterList.java:273)
     at jakarta.mail.internet.ContentDisposition.<init>(ContentDisposition.java:86)
     at jakarta.mail.internet.MimeBodyPart.getDisposition(MimeBodyPart.java:1239)
     at jakarta.mail.internet.MimeBodyPart.getDisposition(MimeBodyPart.java:327)
```

After googling I found that the header is encoded in "Q encoding" (https://en.wikipedia.org/wiki/MIME#Difference_between_Q-encoding_and_quoted-printable) It is also mentioned in RFC2047 (https://www.ietf.org/rfc/rfc2047.txt).
The method jakarta.mail.internet.MimeUtility.decodeText(String etext) is actually able to parse such strings.

**To Reproduce**
See test cases in attached pull request

**Expected behavior**
See test cases in attached pull request

**Envorinment:**
 - Version: 2.0.1

lukasj:
According to [RFC 2231](https://www.rfc-editor.org/rfc/rfc2231) which updates 2047 and explicitly allows using encoding definition in the Content-Disposition header, it is required to inform the client about encoding in parameter value being used through the usage of the `*` in the parameter name (see the definition od _extended-parameter/extended-initial-name_ in the grammar), so in this particular case, the header should look like:

```
filename*==?utf-8?Q?XX...
```
[...]

PR #688, opened 2023-05-10
tests for issue #687

Dr4K4n:


api/src/test/java/jakarta/mail/internet/ParameterListTest.java
```diff
@@ -71,4 +71,33 @@ public void testLongSet() throws Exception {
...
+                             " filename*0*=utf-8''XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX--111111111-XXXXXXXXXXX;\n" +
+                             " filename*1*=XXXXXXXXXXXXXXXXXXXXXXXXXXX.pdf";
+        ParameterList parameterList = new ParameterList(disposition);
+        assertEquals("XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX--111111111-XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX.pdf", parameterList.get("filename"));
+    }
+
+    @Test
+    public void testQEncoding() throws Exception {
+        String disposition = ";\n" +
+                             " filename==?utf-8?Q?XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX--111111111-XXXXXXXXXXXXXXXXXXX?=\n" +
```
lukasj, line 87: `filename==?utf-8?Q?X...` is invalid - it is missing `*`, it should be `filename*==?utf-8?Q?X...`

api/src/test/java/jakarta/mail/internet/ParameterListTest.java
```diff
@@ -71,4 +71,33 @@ public void testLongSet() throws Exception {
...
+        assertEquals("XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX--111111111-XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX.pdf", parameterList.get("filename"));
+    }
+    
+    @Test
+    public void testQEncodingAndUtf8() throws Exception {
+        String disposition = ";\n" +
+                             " filename==?utf-8?Q?XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX--111111111-XXXXXXXXXXXXXXXXXXX?=\n" +
+                             " =?utf-8?Q?XXXXXXXXXXXXXXXXXXX=2Epdf?=;\n" +
+                             " filename*0*=utf-8''XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX--111111111-XXXXXXXXXXX;\n" +
+                             " filename*1*=XXXXXXXXXXXXXXXXXXXXXXXXXXX.pdf";
```
lukasj, line 99: this could probably be handled in a way that if there is a valid variant of the parameter then ignore (and just log) the invalid one in non-string operation mode or through some new, different property.

PR #695, opened 2023-07-22
SharedFileInputStream should comply with spec

### LINKED ISSUE #694: Master SharedFileInputStream closes unexpectedly
**Describe the bug**
Master `SharedFileInputStream` closes unexpectedly after GC even if some slave `SharedFileInputStream`(constructed by `newStream()`) are still being referenced.

Detailed exception is `IOException: Stream Closed`

**To Reproduce**

Here's a simple failing test case to reproduce the error.

```java
@Test
void test() throws Exception {
    // you can use any file to test
    File file = Paths.get("src/test/resources/test_file.jpeg").toFile();
   
    InputStream slaveSharedFileIs = new SharedFileInputStream(file).newStream(0, -1);
   
    System.gc();

    Assertions.assertDoesNotThrow(() -> slaveSharedFileIs.read());
}

```

I guess master `SharedFileInputStream` which is not referenced anymore calls `finalize()` and `close()` when GC works, so slave `SharedFileInputStream` also cannot access the file after that.

To prove, the test case below succeeds.

```java
@Test
void test() throws Exception {
    // you can use any file to test
    File file = Paths.get("src/test/resources/test_file.jpeg").toFile();

    SharedFileInputStream masterSharedFileIs = new SharedFileInputStream(file);
    
    InputStream slaveSharedFileIs = masterSharedFileIs.newStream(0, -1);
   
    System.gc();

    Assertions.assertDoesNotThrow(() -> slaveSharedFileIs.read());
}
```

**Expected behavior**
`RandomAccessFile` should not be closed when there is at least one of remaining `SharedFileInputStream` still being referenced(used)

SeongEon-Jo:
resolves #694 

Added reference to master `SharedFileInputStream` to slaves.

By doing this, I guess it can retain reference to master `SharedFileInputStream` in order not to be junked by garbage collection until the reference to the last derived stream is gone.

Feel free to comment if there are some points to add or fix.

### DIFF
--- a/api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
+++ b/api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
@@ -79,18 +79,6 @@ public class SharedFileInputStream extends BufferedInputStream
     protected long datalen;
 
     /**
-     * True if this is a top level stream created directly by "new".
-     * False if this is a derived stream created by newStream.
-     */
-    private boolean root = true;
-
-    /**
-     * Reference to the top level stream in order not to be junked by garbage collection when there is a remaining derived stream.
-     * null if this is a top level stream
-     */
-    private SharedFileInputStream rootInputStream = null;
-
-    /**
      * A shared class that keeps track of the references
      * to a particular file so it can be closed when the
      * last reference is gone.
@@ -115,20 +103,6 @@ public class SharedFileInputStream extends BufferedInputStream
         public synchronized void close() throws IOException {
             if (cnt > 0 && --cnt <= 0)
                 in.close();
-        }
-
-        public synchronized void forceClose() throws IOException {
-            if (cnt > 0) {
-                // normal case, close exceptions propagated
-                cnt = 0;
-                in.close();
-            } else {
-                // should already be closed, ignore exception
-                try {
-                    in.close();
-                } catch (IOException ioex) {
-                }
-            }
         }
 
         @Override
@@ -217,11 +191,9 @@ public class SharedFileInputStream extends BufferedInputStream
     /**
      * Used internally by the <code>newStream</code> method.
      */
-    private SharedFileInputStream(SharedFileInputStream rootInputStream, SharedFile sf, long start, long len,
+    private SharedFileInputStream(SharedFile sf, long start, long len,
                                   int bufsize) {
         super(null);
-        this.rootInputStream = rootInputStream;
-        this.root = false;
         this.sf = sf;
         this.in = sf.open();
         this.start = start;
@@ -472,15 +444,11 @@ public class SharedFileInputStream extends BufferedInputStream
         if (in == null)
             return;
         try {
-            if (root)
-                sf.forceClose();
-            else
-                sf.close();
+            sf.close();
         } finally {
             sf = null;
             in = null;
             buf = null;
-            rootInputStream = null;
         }
     }
 
@@ -520,14 +488,7 @@ public class SharedFileInputStream extends BufferedInputStream
         if (end == -1)
             end = datalen;
 
-        SharedFileInputStream rootIs;
-
-        if (this.root)
-            rootIs = this;
-        else
-            rootIs = this.rootInputStream;
-
-        return new SharedFileInputStream(rootIs, sf,
+        return new SharedFileInputStream(sf,
                 this.start + start, end - start, bufsize);
     }
 
--- a/api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
+++ b/api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
@@ -85,6 +85,12 @@ public class SharedFileInputStream extends BufferedInputStream
     private boolean master = true;
 
     /**
+     * Reference to the top level stream in order not to be junked by garbage collection when there is a remaining derived stream.
+     * null if this is a top level stream
+     */
+    private SharedFileInputStream masterInputStream = null;
+
+    /**
      * A shared class that keeps track of the references
      * to a particular file so it can be closed when the
      * last reference is gone.
@@ -211,9 +217,10 @@ public class SharedFileInputStream extends BufferedInputStream
     /**
      * Used internally by the <code>newStream</code> method.
      */
-    private SharedFileInputStream(SharedFile sf, long start, long len,
+    private SharedFileInputStream(SharedFileInputStream masterInputStream, SharedFile sf, long start, long len,
                                   int bufsize) {
         super(null);
+        this.masterInputStream = masterInputStream;
         this.master = false;
         this.sf = sf;
         this.in = sf.open();
@@ -511,7 +518,15 @@ public class SharedFileInputStream extends BufferedInputStream
             throw new IllegalArgumentException("start < 0");
         if (end == -1)
             end = datalen;
-        return new SharedFileInputStream(sf,
+
+        SharedFileInputStream masterIs;
+
+        if (this.master)
+            masterIs = this;
+        else
+            masterIs = this.masterInputStream;
+
+        return new SharedFileInputStream(masterIs, sf,
                 this.start + start, end - start, bufsize);
     }
 
--- a/api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
+++ b/api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
@@ -82,13 +82,13 @@ public class SharedFileInputStream extends BufferedInputStream
      * True if this is a top level stream created directly by "new".
      * False if this is a derived stream created by newStream.
      */
-    private boolean master = true;
+    private boolean root = true;
 
     /**
      * Reference to the top level stream in order not to be junked by garbage collection when there is a remaining derived stream.
      * null if this is a top level stream
      */
-    private SharedFileInputStream masterInputStream = null;
+    private SharedFileInputStream rootInputStream = null;
 
     /**
      * A shared class that keeps track of the references
@@ -217,11 +217,11 @@ public class SharedFileInputStream extends BufferedInputStream
     /**
      * Used internally by the <code>newStream</code> method.
      */
-    private SharedFileInputStream(SharedFileInputStream masterInputStream, SharedFile sf, long start, long len,
+    private SharedFileInputStream(SharedFileInputStream rootInputStream, SharedFile sf, long start, long len,
                                   int bufsize) {
         super(null);
-        this.masterInputStream = masterInputStream;
-        this.master = false;
+        this.rootInputStream = rootInputStream;
+        this.root = false;
         this.sf = sf;
         this.in = sf.open();
         this.start = start;
@@ -472,7 +472,7 @@ public class SharedFileInputStream extends BufferedInputStream
         if (in == null)
             return;
         try {
-            if (master)
+            if (root)
                 sf.forceClose();
             else
                 sf.close();
@@ -519,14 +519,14 @@ public class SharedFileInputStream extends BufferedInputStream
         if (end == -1)
             end = datalen;
 
-        SharedFileInputStream masterIs;
+        SharedFileInputStream rootIs;
 
-        if (this.master)
-            masterIs = this;
+        if (this.root)
+            rootIs = this;
         else
-            masterIs = this.masterInputStream;
+            rootIs = this.rootInputStream;
 
-        return new SharedFileInputStream(masterIs, sf,
+        return new SharedFileInputStream(rootIs, sf,
                 this.start + start, end - start, bufsize);
     }
 
--- a/api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
+++ b/api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
@@ -35,15 +35,6 @@ import java.lang.ref.Reference;
  * A <code>RandomAccessFile</code> object is used to
  * access the file data. <p>
  *
- * Note that when the SharedFileInputStream is closed,
- * all streams created with the <code>newStream</code>
- * method are also closed.  This allows the creator of the
- * SharedFileInputStream object to control access to the
- * underlying file and ensure that it is closed when
- * needed, to avoid leaking file descriptors.  Note also
- * that this behavior contradicts the requirements of
- * SharedInputStream and may change in a future release.
- *
  * @author Bill Shannon
  * @since JavaMail 1.4
  */
--- a/api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
+++ b/api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
@@ -23,6 +23,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.RandomAccessFile;
+import java.lang.ref.Reference;
 
 /**
  * A <code>SharedFileInputStream</code> is a
@@ -106,7 +107,7 @@ public class SharedFileInputStream extends BufferedInputStream
         }
 
         @Override
-        protected void finalize() throws Throwable {
+        protected synchronized void finalize() throws Throwable {
             try {
                 in.close();
             } finally {
@@ -449,6 +450,7 @@ public class SharedFileInputStream extends BufferedInputStream
             sf = null;
             in = null;
             buf = null;
+            Reference.reachabilityFence(this);
         }
     }
 
@@ -481,15 +483,19 @@ public class SharedFileInputStream extends BufferedInputStream
      */
     @Override
     public synchronized InputStream newStream(long start, long end) {
-        if (in == null)
-            throw new RuntimeException("Stream closed");
-        if (start < 0)
-            throw new IllegalArgumentException("start < 0");
-        if (end == -1)
-            end = datalen;
+        try {
+            if (in == null)
+                throw new RuntimeException("Stream closed");
+            if (start < 0)
+                throw new IllegalArgumentException("start < 0");
+            if (end == -1)
+                end = datalen;
 
-        return new SharedFileInputStream(sf,
-                this.start + start, end - start, bufsize);
+            return new SharedFileInputStream(sf,
+                    this.start + start, end - start, bufsize);
+        } finally {
+            Reference.reachabilityFence(this);
+        }
     }
 
     // for testing...
@@ -514,7 +520,7 @@ public class SharedFileInputStream extends BufferedInputStream
      * Force this stream to close.
      */
     @Override
-    protected void finalize() throws Throwable {
+    protected synchronized void finalize() throws Throwable {
         super.finalize();
         close();
     }
--- a/api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
+++ b/api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
@@ -480,6 +480,7 @@ public class SharedFileInputStream extends BufferedInputStream
             sf = null;
             in = null;
             buf = null;
+            rootInputStream = null;
         }
     }
 
--- a/api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
+++ b/api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
@@ -23,7 +23,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.RandomAccessFile;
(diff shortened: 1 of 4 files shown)


api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
```diff
@@ -211,9 +217,10 @@ private void init(SharedFile sf, int size) throws IOException {
     /**
      * Used internally by the <code>newStream</code> method.
      */
-    private SharedFileInputStream(SharedFile sf, long start, long len,
+    private SharedFileInputStream(SharedFileInputStream masterInputStream, SharedFile sf, long start, long len,
```
jmehrens, line 220: Change `masterInputStream` to `rootInputStream`

api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
```diff
@@ -84,6 +84,12 @@ public class SharedFileInputStream extends BufferedInputStream
      */
     private boolean master = true;
 
+    /**
+     * Reference to the top level stream in order not to be junked by garbage collection when there is a remaining derived stream.
+     * null if this is a top level stream
+     */
+    private SharedFileInputStream masterInputStream = null;
```
jmehrens, line 91: If this.close() is invoked I would think that the reference to top level stream should be set to null but not closed.

jmehrens (changes requested):
@lukasj

jmehrens (changes requested):
@lukasj The main issue with this change is that there is a finalizer declared.  Should the change really be to just remove the finalizer method as a slightly incompatible change?  Or do we need to convert that code to the Cleaner API?

api/src/main/java/jakarta/mail/util/SharedFileInputStream.java
SeongEon-Jo: You are talking about when the derived stream is closed, Right?

Totally agree with that.

I understand you suggest I set `this.masterInputStream` as `null` when `this.close()` of derived stream is invoked but do not let it close `RandomAccessFile`.

But I think when `this.close()` of the derived stream is invoked, the `RandomAccessFile` is not closed because `SharedFile.close()` is already handling it.

If the stream is not master, `this.close()` invokes `SharedFile.close()`. 

`SharedFile.close()` is like below

```
public synchronized void close() throws IOException {
	    if (cnt > 0 && --cnt <= 0)
		in.close();
	}
```

As you can see, It does not close `RandomAccessFile` except when `cnt` is 1, which means it is the last object to have an access to the `SharedFile` or `RandomAccessFile`.

So, I guess I only need to add a code to set `this.masterInputStream`as `null` when `this.close()` of derived stream is invoked.

Anything wrong I misunderstood?

jmehrens: Correct.  Essentially we can just unconditionally null this.masterInputStream as the last statement:

````
 @Override
    public void close() throws IOException {
        if (in == null)
            return;
        try {
            if (master)
                sf.forceClose();
            else
                sf.close();
        } finally {
            sf = null;
            in = null;
            buf = null;
            masterInputStream = null;
        }
    }
````

I'm thinking that if we are pinning the root stream by holding a strong reference on creation of a child stream then when the child is closed the link from child SharedFileInputStream to the root SharedFileInputStream should be unlinked (set to null).  Not that we are leaking a file handle but just that we are retaining a small amount of garbage if the child stream is held on to by the caller (3rd party code).

Comment (SeongEon-Jo) [reply api/src/main/java/jakarta/mail/util/SharedFileInputStream.java:91]: Got it. I misunderstood you were saying my code causes the derived stream to close the `RandonAccessFile`   😂

I will edit `this.close()` like you suggested. It is exactly the same way I was about to do 😎

**Comment by jmehrens** [comment]
@SeongEon-Jo You'll have to include a test case with this PR.  Modify the hardcoded file path to use File.createTempFile.

I'm thinking this test will fail even under the new patch.

````
@Test
void testGrandchild() throws Exception {
    // you can use any file to test
    File file = Paths.get("src/test/resources/test_file.jpeg").toFile();
   
    InputStream grandChild = new SharedFileInputStream(file).newStream(0, -1).newStream(0, -1);
   
    System.gc();

    Assertions.assertDoesNotThrow(() ->grandChild.read());
}
````

**Comment by SeongEon-Jo** [comment]
I added a test case but slightly changed the code you gave because `SharedFileInputStream.newStream()` returns `InputStream`, not `SharedInputStream`.

So I just casted it to `SharedFileInputStream` in order to invoke `newStream()`.

Test result was "passed".

Also, I ran the same test in junit5 like below.

```
@Test
    void testGrandchild() throws Exception {
        File sampleFile = File.createTempFile("test", "test");

        try (InputStream grandChild = ((SharedFileInputStream) new SharedFileInputStream(sampleFile).newStream(0, -1)).newStream(0, -1)) {
            System.gc();
            Assertions.assertDoesNotThrow(() -> grandChild.read());
        }
    }
```

Result was also "passed".

I only added a case of grand child. But if you want me to add another cases, feel free to tell me about that.

**Comment by jmehrens** [comment]
> Result was also "passed".
> 
> I only added a case of grand child. But if you want me to add another cases, feel free to tell me about that.

Oh I see.  The reason this passes is that the un-reference parent gets garbage collected but the call to close by the finializer is a no-op because of the reference counting.  All good then.

You should add your original test case too as a method to the JUnit test.  You should copy one of the license headers from the other test cases.

**Comment by SeongEon-Jo** [comment]
> You should add your original test case too as a method to the JUnit test. You should copy one of the license headers from the other test cases.

Added the license header and child test case.

But I want to notify that in the version before this patch, the test results were different depending on whether I invoke `Assertions.assertDoesNotThrow()` of junit5 or just invoke `child.read()`.

For example, in a test code I gave in https://github.com/jakartaee/mail-api/issues/694,

```
@Test
void test() throws Exception {
    // you can use any file to test
    File file = Paths.get("src/test/resources/test_file.jpeg").toFile();
   
    InputStream slaveSharedFileIs = new SharedFileInputStream(file).newStream(0, -1);
   
    System.gc();

    Assertions.assertDoesNotThrow(() -> slaveSharedFileIs.read());
}
```
it failed, which means `slaveShareFiles.read()` throws and `IOException`.

But when I tested with the code below,

```
@Test
void test() throws Exception {
    // you can use any file to test
    File file = Paths.get("src/test/resources/test_file.jpeg").toFile();
   
    InputStream slaveSharedFileIs = new SharedFileInputStream(file).newStream(0, -1);
   
    System.gc();

    slaveSharedFileIs.read();
}
```
it passed, which means `slaveSharedFileIs.read()` does not throw `IOException`

I guess this occurs because of the code scope but not sure of the exact reason.

Of course, after my patch, no matter whether I invoke `Assertions.assertDoesNotThrow()` or just invoke `child.read()`, all test succeeded, not throwing `IOException`.

The reason I am notifying this is because I am not that sure if the test code I've added properly covers all the problematic situations.

Please let me know your opinion and suggest test cases I can add to cover them.

**Comment by jmehrens** [comment]
> But I want to notify that in the version before this patch, the test results were different depending on whether I invoke `Assertions.assertDoesNotThrow()` of junit5 or just invoke `child.read()`.

Put a comment on this behavior so future maintainers don't try to refactor and break the test.

File: api/src/test/java/jakarta/mail/util/SharedFileInputStreamTest.java [hunk 3]
```diff
@@ -0,0 +1,53 @@
...
+    @Test
+    public void testGrandChild() throws Exception {
+        File file = File.createTempFile("test", "test");
+
+        try (InputStream grandChild = ((SharedFileInputStream) new SharedFileInputStream(file).newStream(0, -1)).newStream(0, -1)) {
+            System.gc();
+            grandChild.read();
+        } catch (IOException e) {
+            fail("IOException is not expected");
+        }
```
jmehrens, line 51: Should add a finally block which invokes delete on the file.

api/src/test/java/jakarta/mail/util/SharedFileInputStreamTest.java
```diff
@@ -0,0 +1,53 @@
...
+    @Test
+    public void testChild() throws Exception {
+        File file = File.createTempFile("test", "test");
+
+        try (InputStream childStream = new SharedFileInputStream(file).newStream(0, -1)) {
+            System.gc();
+            childStream.read();
+        } catch (IOException e) {
+            fail("IOException is not expected");
+        }
```
jmehrens, line 38: Should add a finally block which invokes delete on the file.

api/src/test/java/jakarta/mail/util/SharedFileInputStreamTest.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+
+    @Test
+    public void testGrandChild() throws Exception {
+        File file = File.createTempFile("test", "test");
```
jmehrens, line 44:
```suggestion
        File file = File.createTempFile(SharedFileInputStreamTest.class.getName(), "testGrandChild");
```

jmehrens (changes requested):

SeongEon-Jo:
> Put a comment on this behavior so future maintainers don't try to refactor and break the test.

Added a comment to the test class.

Feel free to tell me if there is something to add or edit in the comment.

jmehrens:
@SeongEon-Jo  I've been looking at the source code and I think the original behavior of this class is by design. Per:

> Note that when the SharedFileInputStream is closed,
> all streams created with the <code>newStream</code>
> method are also closed.  This allows the creator of the
> SharedFileInputStream object to control access to the
> underlying file and ensure that it is closed when
> needed, to avoid leaking file descriptors.  Note also
> that this behavior contradicts the requirements of
> SharedInputStream and may change in a future release.

That is why there are 2 code paths for closing and force closing.  So in theory this issue could be closed as will not fix because the behavior is actually documented.  If we were going to change the behavior so this didn't happen then we could instead pinning the root stream we could just reference count all streams instead of treating one stream as the root marked by a boolean.  Meaning we never call forceClose.

Other than the test case provided, how did you come across this issue in the wild?  Is it the case of some abstraction returning a sub stream?

SeongEon-Jo:
@jmehrens

As I understand, the document you mentioned is about that closing the root stream will close any derived streams created by `newStream()`.

It is a sensible behavior and I totally agree with that, and that's not what I am talking about.

What I am pointing out in the issue (https://github.com/jakartaee/mail-api/issues/694) and this PR is that G.C may close the root `SharedFileInputStream` unexpectedly when there is nothing to keep a reference to it.

The solution I try to give is also about making the root `SharedFileInputStream` not be junked by G.C by pinning it as the way derived streams retain its reference.

I guess a way of the reference counting you suggested seems to solve the problem only after you remove its `finalize()` method because if G.C invokes the root `SharedFileInputStream`'s `finalize()`, it will make it close its `RandomAccessFile` whatever the reference count is.

Actually, I also initially thought of the reference counting as a solution, but just made derived streams keep a reference to the root stream because removing or modifying `finalize()` is quite burdensome 😂

SeongEon-Jo:
> Other than the test case provided, how did you come across this issue in the wild? Is it the case of some abstraction returning a sub stream?

I found it when I make a `MimeMessage` with `SharedFileInputStream` and extract some of the `DataHandler` in it and get `DataSource` and also `InputStream` in it.

Let me give you a sample code to help you understand.

Firstly, make a new `MimeMessage` with `SharedFileInputStream` and extract some of the `InputStream` of  a`DataSource` of a `DataHandler`.

```java
public List<InputStream> extract() {

        MimeMessage message = new MimeMessage(session, new SharedFileInputStream(getFile()));

        List<InputStream> results = new ArrayList<>();

        if (...) {
            InputStream extractedInputStream = message.getDataHandler().getDataSource().getInputStream();
            results.add(extractedInputStream);
        }
        return results;
    }
```

Note that `MimeMessage` actually has a `SharedFileInputStream` created by `newStream()` and there's no reference to a root stream .

https://github.com/jakartaee/mail-api/blob/29fc0622412712822f27c4303cba4a96aea32100/api/src/main/java/jakarta/mail/internet/MimeMessage.java#L335-L356 

Secondly, invoke the `extract()` and do something with the returned `InputStream`, getting size for example.

```java
public List<Integer> getSize() {
        List<InputStream> extracted = extract();

        return extracted.stream()
                .map(is -> is.available())
                .collect(Collectors.toList());
    }
```

I found `is.available()` threw `IOException: Stream Closed`.

It's because G.C invoked `finalize()` of a root stream somewhen between `extract()` and `getSize()`, closing the `RandomAccessFile`.

As a result, extracted `InputStream` which was also made of `SharedFileInputStream` that `MimeMessage` has, also became unavailable.

I also checked `RandomAccessFile` is closed while running `getSize()` by debugger.

스크린샷 2023-08-07 오전 10 17 10

jmehrens:
> I guess a way of the reference counting you suggested seems to solve the problem only after you remove its `finalize()` method because if G.C invokes the root `SharedFileInputStream`'s `finalize()`, it will make it close its `RandomAccessFile` whatever the reference count is.

I think the problem is this class is designed it such a way that it is a candidate for [java.lang.ref.Reference.reachabilityFence](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/ref/Reference.html#reachabilityFence-java.lang.Object-)

I would think that we don't need to pin the root stream.  It just needs to be kept alive long enough such that the counter gets incremented before the root is freed.  E.G:
````
public synchronized InputStream newStream(long start, long end) {
    try {
            if (in == null)
                throw new RuntimeException("Stream closed");
            if (start < 0)
                throw new IllegalArgumentException("start < 0");
            if (end == -1)
                end = datalen;
            return new SharedFileInputStream(sf,
                    this.start + start, end - start, bufsize);
    } finally {
         Reference.reachabilityFence(this);
    }
}
````

SeongEon-Jo:
@jmehrens 

I did same tests (https://github.com/jakartaee/mail-api/pull/695#issuecomment-1664174805) based on the sample code you gave. (Note that tests were run **with original `SharedFileInputStream` before my patch.**)

Firstly, added `Reference.reachabilityFence(this);` at the `finally` scope of `newStream()`.

Secondly, added `synchronized` at the `close()` without adding `Reference.reachabilityFence(this);` at the `finally` scope of `newStream()`.

Third, added both.

Test all failed.

Actually, I don't understand what `synchronized` has to do with G.C. Is it just for supporting synchronizing in multi-thread?

If there is something I can refer to, please let me know. so that I can add another tests.

> My preference would be to use the Cleaner API on the SharedFile and get rid of the finalize method on the SharedFileInputStream.

This seems to be one of the reasonable ways to deal with this problem because it leaves file's lifecycle to the `SharedFile` that actually references to it, rather than to the `SharedFileInputStream`.

I also found that just removing `finalize()` of `SharedFileInputStream` made it all test passed.

I think if you fix the problem in that way, a completely new work needs to be done, closing this PR.

It seems to be not that hard work. Do you want me to work on it?

jmehrens:
> Test all failed.
I think this because we also have to get rid of the concept of the root stream being able to close all streams and simply reference count. As in get rid of root/master boolean and forceClose methods.  I could be wrong though.

> Actually, I don't understand what `synchronized` has to do with G.C. 
In this case synchronized modifier means that the current thread has locked 'this' for the whole method call.  Meaning that 'this' is reachable during the whole method call.  Without synchronized modifier `this` can be G.C during the method call if there are no more references to `this` where `this` is the root stream in:

`InputStream slaveSharedFileIs = new SharedFileInputStream(file).newStream(0, -1);`

This is explained at the bottom of:
[java.lang.ref.Reference.reachabilityFence](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/ref/Reference.html#reachabilityFence-java.lang.Object-)

>Is it just for supporting synchronizing in multi-thread? 
Usually it is to ensure that a close stream is seen as closed to all threads.


> > My preference would be to use the Cleaner API on the SharedFile and get rid of the finalize method on the SharedFileInputStream.
> 
> This seems to be one of the reasonable ways to deal with this problem because it leaves file's lifecycle to the `SharedFile` that actually references to it, rather than to the `SharedFileInputStream`.
> 
> I also found that just removing `finalize()` of `SharedFileInputStream` made it all test passed.
> 
> I think if you fix the problem in that way, a completely new work needs to be done, closing this PR.
> 
> It seems to be not that hard work. Do you want me to work on it?

Let me try to get consensus on the approach.

The more I think about it the reason the finalizer exists on the SharedFileInputStream is to make the reference counting accurate.  Which enables the last stream to close the native resource.  Which is important if the last/root stream is strongly referenced.

SeongEon-Jo:
> I think this because we also have to get rid of the concept of the root stream being able to close all streams and simply reference count. As in get rid of root/master boolean and forceClose methods. I could be wrong though.

I see. All test passed after getting rid of master boolean and `forceClose()`. Please be aware of this result and refer to it for future works.

> Let me try to get consensus on the approach.

All right. Please let me know if there's any progress and anything that I can help.

jmehrens:
@SeongEon-Jo Apologizes for not getting back to you sooner on this PR.  After looking this over and what has been learned from your previous code I think the approach should be:

1. SharedFileInputStream::close should use [java.lang.ref.Reference.reachabilityFence](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/ref/Reference.html#reachabilityFence-java.lang.Object-) since it is not synchronized.  I don't want to mark this method as synchronized at this time because I think Bill intentionally didn't mark it as synchronized.  That issue can be handled in another ticket.
2.  SharedFileInputStream::newStream should use [java.lang.ref.Reference.reachabilityFence](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/ref/Reference.html#reachabilityFence-java.lang.Object-).  This is not needed for correctness but shows the intent that we want to make sure the parent is kept alive.
3. Root/master boolean field should be deleted along with references to it.
4. forceClose() method should be removed along with references to it.
5. API documentation should be updated to remove the paragraph about the root stream closing all streams.
6. Both finalizers should be marked as synchronized.
7. We should avoid using the Cleaner API at this time because it is going to cause more work for this PR.

I'm thinking with these changes this class should work as you expect it to and we won't need to pin the root stream.

Thanks for all your efforts on this PR and helping me gain an understanding of what is going on with this class.

SeongEon-Jo:
@jmehrens 

All sounds good in your approach.

All test passed fortunately after applying your suggestions.

> Thanks for all your efforts on this PR and helping me gain an understanding of what is going on with this class.

You're welcome. I'm happy to contribute. 😎

Feel free again to let me know if  you have more to add or edit.

jmehrens:
@SeongEon-Jo Looks like we are targeting JDK8 still in the pom.xml.  Instead of calling [reachabilityFence](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/ref/Reference.html#reachabilityFence-java.lang.Object-)

You can do the same with:
````
import java.util.Objects;

Objects.requireNonNull(this); //TODO: replace with Reference.reachabilityFence
````

jmehrens:
@SeongEon-Jo Sorry for the delay on this.  Codewise, I think this is all good.  Just need to do a review of how it is used in the codebase before I merge it.

I think you need to update the https://github.com/jakartaee/mail-api/blob/master/doc/release/CHANGES.txt
with something like:

````
CHANGES IN THE 2.1.3 RELEASE
          ----------------------------
E 695  SharedFileInputStream should comply with spec
````

I think you also need to put a compatibility note in: https://github.com/jakartaee/mail-api/blob/master/www/docs/COMPAT.txt
for the 2.1.3 RELEASE.  Something like:

````
- SharedFileInputStream should comply with spec
    The root SharedFileInputStream no longer closes all streams created with the newStream.  This behavior was not compliant with the contract specified in the SharedInputStream interface which specifies that all streams must be closed before the shared resource is closed.
````

I finally found in my notes the explanation on requireNonNull works: https://mail.openjdk.org/pipermail/core-libs-dev/2018-February/051312.html

Thanks for the help! Nice work!

jmehrens:
@SeongEon-Jo I haven't forgotten about this!  I'll do my best to get this reviewed and merged.  Thank you for your patience.

jmehrens:
Committed as:
https://github.com/jakartaee/mail-api/commit/6c9ac509c4b515a15dbb790309bc8469c911d308

PR #701, opened 2023-11-12
Jakarta Mail erroneously assumes that classes can be loaded from Thread#getContextClassLoader

jmehrens:
https://github.com/jakartaee/mail-api/issues/665

PR #702, opened 2023-11-27
Proposed workaround to the jakarta.mail.util.StreamProvider

vionta:
Currently, when the instantiation of the StreamProvider Implementation fails using the Thread Context Class Loader, there's no fallback to the newInstance on the App Class Loader. 
This is a proposed workaround, that allows to load the StreamProvider implementation class indicating the explicit class name (Declaration example under this line):
System.setProperty("jakarta.mail.util.StreamProvider","org.eclipse.angus.mail.util.MailStreamProvider");


jmehrens:
I think I have this case covered in https://github.com/jakartaee/mail-api/pull/701

vionta:
Hi, I made a specific test on the environment and code from https://github.com/jakartaee/mail-api/pull/701 seems to work.

jmehrens:
@vionta That is great news!  I'm holding off on merging #701 until I complete the matching PR for Activation.

Issue #710, opened 2024-02-15
Cannot parse messages without a session


jmehrens:
Move of https://github.com/eclipse-ee4j/angus-mail/issues/121

PR #711, opened 2024-02-16
Cannot parse messages without a session #710

jmehrens:
Signed-off-by: jmehrens jason_mehrens@hotmail.com

### DIFF
--- a/api/src/main/java/jakarta/mail/internet/MimeMessage.java
+++ b/api/src/main/java/jakarta/mail/internet/MimeMessage.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2023 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2024 Oracle and/or its affiliates. All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


jmehrens:
Per @bshannon (R.I.P) on [What is the use of javax.mail.Session?](https://stackoverflow.com/questions/22657918/what-is-the-use-of-javax-mail-session):

>To add to that... The support for a null Session is there mostly to provide some reasonable behavior in the case where people screw up and fail to create a Session.

Classic!  Point being is that null Session is allowed in MimeMessage the fallback is to use the default session when needed.  The approach to this patch is to mimic the behavior in Transport.send0 which is:

````
Session s = (msg.session != null) ? msg.session :
                Session.getDefaultInstance(System.getProperties(), null);
````
The alternate patch would be to call StreamProvider.provider() in case the session is null.  I think this is more consistent with what is already done.

lukasj (approved):

lukasj:
@jmehrens I'll respin 2.1.3 so this gets there. Do you have anything else for 2.1.3 before I do that?

Issue #590, opened 2021-11-24
move from MailDateFormat (not thread safe) to DateTimeFormatter


valenpo:
With high load and usage of MimeMessage plenty of locks due to MailDateFormat. Maybe it is time move to something thread safe like DateTimeFormatter?

jmehrens:
MailDateFormatter has lots of exceptions for parsing that make the move to DateTimeFormatter not viable.

I think the best fix here is to remove the lock and static field.  Then have each call create a method local instance.

Thoughts @anthonyvdotbe?

anthonyvdotbe:
@jmehrens yes, I agree that your proposal is the best option here

PR #713, opened 2024-02-28
Move from MailDateFormat (not thread safe) to DateTimeFormatter #590

jbescos:
https://github.com/jakartaee/mail-api/issues/590

I only saw references in `MimeMessage`. I updated that class to not use `MailDateFormat`.


api/src/main/java/jakarta/mail/internet/MimeMessage.java
```diff
@@ -902,10 +909,8 @@ public Date getSentDate() throws MessagingException {
         String s = getHeader("Date", null);
         if (s != null) {
             try {
-                synchronized (mailDateFormat) {
-                    return mailDateFormat.parse(s);
-                }
-            } catch (ParseException pex) {
+                return Date.from(Instant.from(mailDateFormat.parse(s)));
```
jmehrens, line 912: MailDateFormat has a concept of isLenient and per the API documents it was not clear to me if the default is lenient or strict.  I wonder if you need to use withResolverStyle to set that behavior and how compatible with existing behavior, mainly with parsing of dates.

Looking at the [MailDateFormatTest](https://github.com/jakartaee/mail-api/blob/master/api/src/test/java/jakarta/mail/internet/MailDateFormatTest.java) there are some tests that seem to imply leniency with lots of exceptions:

1. lenientMustBeBackwardCompatible()
2. mustUseLeniencyForParsing()
3. mustFailOrSkipCfws()
4. lenientMustAcceptNonMilitaryZoneNames()

I would think that for testing this version of the patch you would need to port the MailDateFormatTest into the MimeMessageTest.  Which seems like a lot of work.

What are your thoughts on just getting rid of the static field and the synchronized block and just creating some method local garbage by just creating a new MailDateFormat for parsing and formatting?

jbescos:
I don't see a feasible way to replace MailDateFormat, and it has many specifics. Probably moving to DateTimeFormatter is not the way to go.

Issue #709, opened 2024-02-15
Support minimum JDK of 11+ and minimum of Android API 33+


jmehrens:
Per [Jakarta EE Platform 10](https://jakarta.ee/specifications/platform/10/):
>Minimum Java SE Version
>Java SE 11 or higher

This means that mail-api would no longer be building for JDK 8.  Moving an Android API version 26+ or greater will give access to `java.lang.invoke`, `java.util.Base64`, and `java.time`.  However, all versions prior to Android 13 (API level 33) will be deprecated so we must assume API level 33 Android is required.  API 26+ may be able to run with mail-api but, is not supported and we will close any tickets that are not meeting the API requirements.

Per [Meet Google Play's target API level requirement](https://developer.android.com/google/play/requirements/target-sdk):

> Starting on August 31, 2023:
>
>    New apps must target Android 13 (API level 33) or higher; except for [Wear OS](https://developer.android.com/wear) apps, which must target a version between Android 11 (API level 30) and Android 13 (API level 33), inclusive.
>
>  App updates must target Android 13 or higher and adjust for behavioral changes in Android 13; except for Wear OS apps, which must target Android 11.
>
>Permanently private apps—which are restricted to users in a specific organization and are intended for internal distribution only—aren't required to meet the target API level requirements.

[snip]

>...Furthermore, as of Android 10 (API level 29), users [see a warning](https://developer.android.com/about/versions/10/behavior-changes-all#low-target-sdk-warnings) when they start an app for the first time if the app targets Android 5.1 (API level 22) or lower.

Per [Which Java APIs can I use in my Java or Kotlin source code?](https://developer.android.com/build/jdks) the API level 32 and API level 33 are versions that support Java 11 source syntax.  API level 34 supports Java 17 source syntax. 

The [Android API Levels](https://apilevels.com/) site shows the usage of devices.

/cc @lukasj

lukasj:
for 2.1.x/EE10 branch (will be created from master once 2.1.3 gets out) I'd keep it the way it is - the change from 8 could be perceived as incompatible by some users; but we can still say 11 is supported and since 2.1.3 support for 8 is limited and provided on the best effort basis only (if someone asks).

after branching, the master (API 2.2.x) will be moved to 11+ only

lukasj:
> If semver is followed to the letter 

No, it is not, updating min SE level is not being considered as a breaking change by EE (AFAIK)

aalmiray:
[...]

PR #714, opened 2024-02-29
Support minimum JDK of 11+ and minimum of Android API 33+ #709

jbescos:
https://github.com/jakartaee/mail-api/issues/709

I set it as a draft for now, as we have to wait for branch 2.2.x and I will need to check also some documents.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -83,6 +83,8 @@
 
         <!-- version identifier in jakarta.mail.Version -->
         <mail.version>${project.version}</mail.version>
+        <maven.compiler.release>11</maven.compiler.release>
+        <maven.compiler.testRelease>11</maven.compiler.testRelease>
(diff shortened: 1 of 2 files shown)


api/pom.xml
```diff
@@ -379,7 +365,7 @@
                           osgi.serviceloader;filter:="(osgi.serviceloader=jakarta.mail.util.StreamProvider)";
                             osgi.serviceloader="jakarta.mail.util.StreamProvider";
                               cardinality:=multiple;resolution:=optional,
-                          osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
+                          osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=11))"
```
lukasj, line 368: this line can be removed (used to be a workaround)

Issue #699, opened 2023-11-03
Multipart performs blocking call in every instantiation


jonathannaguin:
**Describe the bug**
When a `Multipart` object is created, the field `streamProvider` is also instantiated which relies ultimately on FactoryFinder#find (https://github.com/jakartaee/mail-api/blob/master/api/src/main/java/jakarta/mail/util/FactoryFinder.java#L35) that performs blocking operations.

**To Reproduce**
Create a new MimeMultipart object, for example
```
new MimeMultipart("mixed");
```
on a project with https://github.com/reactor/BlockHound installed.

**Expected behavior**
Either the FactoryFinder needs to be rewritten to not use non-blocking operations or `Multipart` could use a static reference so only it's blocked at boot up.

jmehrens:
Related: https://github.com/jakartaee/mail-api/issues/101

jbescos:
Hi @jonathannaguin, where is the blocking operation thing in FactoryFinder?, I cannot find it.

jmehrens:
@jbescos I assumed that the blocking was from Class::forName, ServiceLoader::load, etc.

My initial understanding of the fix is to follow advice from ServiceProvider "... Users are recommended to cache the result of this method."  I don't think we are doing this everywhere but, I'm not sure if it is intended or not.

jmehrens:
The super class Multipart has a protected instance field that most likely causing this issue.

jonathannaguin:
That's right @jmehrens , https://github.com/jakartaee/mail-api/blob/master/api/src/main/java/jakarta/mail/Multipart.java#L69 will instantiate a StreamProvider (https://github.com/jakartaee/mail-api/blob/master/api/src/main/java/jakarta/mail/util/StreamProvider.java#L195) which will use FactoryFinder to find and instantiate given provider.

I was able to workaround some of this by setting some system properties so the Factory finds the `classNames` from the properties (avoiding factoryFromServiceLoader).

jmehrens:
I think that field should be deprecated and removed when move to next major release. If we added a protected method instead of a field we could do searching of the provider from the session of parent part and or other parts.

That said, the easy solution is to create a private static final field and use that to assign the protected instance field.  I just have not done the research to figure out if we lose anything by not running this code for every instance.

jbescos:
[...]

PR #716, opened 2024-03-04
Multipart performs blocking call in every instantiation #699

jbescos:
https://github.com/jakartaee/mail-api/issues/699

It will try to get the StreamProvider from the session of the Message.


api/src/main/java/jakarta/mail/Multipart.java
```diff
@@ -266,4 +269,27 @@ public synchronized Part getParent() {
     public synchronized void setParent(Part parent) {
         this.parent = parent;
     }
+
+    protected StreamProvider provider() {
+        if (streamProvider == null) {
+            synchronized (this) {
+                if (streamProvider == null) {
+                    if (parent != null && parent instanceof Message) {
```
jmehrens, line 277: I would assume that this should loop while the parent is not null. If we find a message and it has non-null provider use it and break out.

api/src/main/java/jakarta/mail/Multipart.java
```diff
@@ -266,4 +269,27 @@ public synchronized Part getParent() {
     public synchronized void setParent(Part parent) {
         this.parent = parent;
     }
+
+    protected StreamProvider provider() {
```
jmehrens, line 273: The writeTo and parse methods both throw MessagingException so I think it would make sense to add throws `MessagingException` here.  We can then wrap `ServiceConfigurationError | RuntimeException` in a MessagingException that the caller is ready to receive.

api/src/main/java/jakarta/mail/Message.java
```diff
@@ -705,4 +707,33 @@ protected void setExpunged(boolean expunged) {
...
+     * Obtains the {@link StreamProvider} from the session, if exists.
+     * Otherwise it obtains it from
+     * {@link Session#getDefaultInstance(java.util.Properties, Authenticator)}.
+     *
+     * @return the StreamProvider implementation from the session.
+     * @throws MessagingException if errors.
+     *
+     * @since JavaMail 2.2
+     */
+    protected StreamProvider provider() throws MessagingException {
```
jmehrens, line 721: I used a short name because this method was private.  Now that it is protected maybe it should be `getStreamProvider` to match Session API?

api/src/main/java/jakarta/mail/Message.java
```diff
@@ -705,4 +707,33 @@ protected void setExpunged(boolean expunged) {
...
+        try {
+            try {
+                final Session s = this.session;
+                if (s != null) {
+                    return s.getStreamProvider();
+                } else {
+                    return Session.getDefaultInstance(System.getProperties(),
```
jmehrens, line 728: We don't have to solve this point in this ticket but, you made a good point about default instance being static which means we could consider using `return StreamProvider.provider()`.  My original thinking was to use the default session to match what was done in Transport and to avoid paying the lookup over and over.

jmehrens (changes requested):

api/src/main/java/jakarta/mail/Message.java
jbescos: I was re-thinking it and what makes more sense to me is:

If the user wants to use its custom StreamProvider, he has to pass the Session in the Message. If he does not pass the session is returns what the default session instance has.

Otherwise it will suffer the performance.

jbescos:
@jmehrens I am redesigning it, because I found more usages of `StreamProvider`. I moved the `getStreamProvider` to the `Part` interface.

api/src/main/java/jakarta/mail/Multipart.java
jbescos: It should be in place now. Now the parent implements #getStreamProvider so it will deal with that.

api/src/main/java/jakarta/mail/Multipart.java
jbescos: Done

api/src/main/java/jakarta/mail/Message.java
jbescos: Yes, I was thinking the same. It is named `getStreamProvider` now.

api/src/main/java/jakarta/mail/Multipart.java
jmehrens: I realized I missed a detail in that during parsing the parent may not be attached.  I think we need check DataSource for messageaware and get session (which has an accessor for streamprovider) from messagecontext in the subclasses.

This should help parsing.

jbescos: I am checking this:
https://github.com/jakartaee/mail-api/blob/bf2bfc18c0d2b9b3f57417203c2a462a3a94a7de/api/src/main/java/jakarta/mail/internet/MimeMultipart.java#L261

It looks this is already obtained and set it as a parent. So we can just ask the parent for the `StreamProvider` instance

api/src/main/java/jakarta/mail/Multipart.java
```diff
@@ -266,4 +259,24 @@ public synchronized Part getParent() {
...
+     * @throws MessagingException if errors.
+     *
+     * @since JavaMail 2.2
+     */
+    protected StreamProvider getStreamProvider() throws MessagingException {
+        Part parent = this.parent;
+        if (parent != null) {
+            return parent.getStreamProvider();
+        } else {
+            return Session.getDefaultInstance(System.getProperties(), null).getStreamProvider();
```
jmehrens, line 278: The else needs the nested catches:

````
try {
   try {
       return Session.getDefaultInstance(System.getProperties(), null).getStreamProvider();
    } catch (ServiceConfigurationError sce) {
        throw new IllegalStateException(sce);
    }
} catch (RuntimeException re) {
   throw new MessagingException("Unable to get " + StreamProvider.class.getName(), re);
}
`````

or needs to just invoke the super method.

File: api/src/main/java/jakarta/mail/Part.java [hunk 6]
```diff
@@ -453,4 +454,17 @@ Enumeration<Header> getMatchingHeaders(String[] header_names)
...
+     * Obtains the {@link StreamProvider}.
+     * It defaults to {@link Session#getDefaultInstance(java.util.Properties, Authenticator)}.
+     *
+     * @return the StreamProvider.
+     * @throws MessagingException if errors.
+     *
+     * @since JavaMail 2.2
+     */
+    default StreamProvider getStreamProvider() throws MessagingException {
+        return Session.getDefaultInstance(System.getProperties(), null).getStreamProvider();
```
jmehrens, line 468: Needs the nested catches:
````
try {
   try {
       return Session.getDefaultInstance(System.getProperties(), null).getStreamProvider();
    } catch (ServiceConfigurationError sce) {
        throw new IllegalStateException(sce);
    }
} catch (RuntimeException re) {
   throw new MessagingException("Unable to get " + StreamProvider.class.getName(), re);
}

Also I noticed there is a more specific exception of NoSuchProviderException.  This was originally for Provider implementations so it might be abuse to reuse the more specific exception.  Something to consider.

**Review by jmehrens — CHANGES_REQUESTED** [review]
I like your new approach.  Very clever.

**Comment by jmehrens** [comment]
Need to add entry to https://github.com/jakartaee/mail-api/blob/master/doc/release/CHANGES.txt

**Comment by jbescos** [comment]
> Need to add entry to https://github.com/jakartaee/mail-api/blob/master/doc/release/CHANGES.txt

Done

File: api/src/main/java/jakarta/mail/Part.java (hunk 6 above)
Comment (jbescos) [reply api/src/main/java/jakarta/mail/Part.java]: NoSuchProviderException extends MessagingException, I think we don't need to catch it.

File: api/src/main/java/jakarta/mail/Multipart.java (hunk 5 above)
Comment (jbescos) [reply api/src/main/java/jakarta/mail/Multipart.java]: Multipart does not implement Part. We have to use your approach with the try/catch. It is in place.

**Comment by jbescos** [comment]
Lets wait for @lukasj review too. My impression is this requires the new version 2.2.0, and we will need other PR before to upgrade to that version.

File: api/src/main/java/jakarta/mail/Multipart.java (hunk 5 above)
Comment (jmehrens) [reply api/src/main/java/jakarta/mail/Multipart.java]: Now I need to re-review to understand that mult-part subparts correctly reach session of a message.

File: api/src/main/java/jakarta/mail/Part.java (hunk 6 above)
Comment (jbescos) [reply api/src/main/java/jakarta/mail/Part.java]: I see, it makes sense to me to throw NoSuchProviderException instead of MessagingException. I think it is a good time to do this. I am going to implement it.

Comment (jmehrens) [reply api/src/main/java/jakarta/mail/Part.java]: I was thinking of keeping the signature `StreamProvider getStreamProvider() throws MessagingException` so we could continue to throw other specific exceptions in the future.  We still actually would throw a more specific exception `new NoSuchProviderException` even though the signature was more general.

File: api/src/test/java/jakarta/mail/PartTest.java [hunk 7]
```diff
@@ -0,0 +1,154 @@
...
+        assertEquals(SESSION_1.getStreamProvider(), multipart.getStreamProvider());
+    }
+
+    @Test
+    public void specifySession2() throws MessagingException {
+        Message message = new MimeMessage(SESSION_2);
+        Multipart multipart = new MimeMultipart();
+        message.setContent(multipart);
+        assertEquals(SESSION_2.getStreamProvider(), multipart.getStreamProvider());
+    }
```
jmehrens, line 79: I think you should test that a subpart of a multipart can find the message:

````
@Test
    public void multipartBodyPart() throws Exception {
        MimeMessage m = new MimeMessage(SESSION_2);
        MimeMultipart mmp = new MimeMultipart("mixed");
        MimeBodyPart mbp = new MimeBodyPart();
        mbp.setDisposition(Part.INLINE); //Do we really need an attachment part test too??
        mbp.setText("none");
        mmp.addBodyPart(mbp);
        m.setContent(mmp);
        
        StreamProvider root = SESSION_2.getStreamProvider();
        assertEquals(root, m.getStreamProvider());
        assertEquals(root, mmp.getStreamProvider());
        assertEquals(root, mbp.getStreamProvider());
    }
````

**Review by jmehrens — CHANGES_REQUESTED** [review]

File: api/src/main/java/jakarta/mail/BodyPart.java [hunk 8]
```diff
@@ -39,13 +39,6 @@ public abstract class BodyPart implements Part {
      */
     protected Multipart parent;
 
-    /**
-     * Instance of stream provider.
-     *
-     * @since JavaMail 2.1
-     */
-    protected final StreamProvider streamProvider = StreamProvider.provider();
```
jmehrens, line 47: The more I think about it, `Session.getDefaultInstance(System.getProperties(), null).getStreamProvider();` is going to pin the classloader of the implementation which is not going to play nice when implementation classloader is shutdown.  Which in a sense is a whole new issue.

That said, should we keep this field, remove final, maybe add volatile, and lazy set it to `StreamProvider.provider();` when we need to?  We shouldn't even need any locking as the race is benign.

jmehrens (changes requested):

api/src/main/java/jakarta/mail/BodyPart.java
jbescos: I would like to be consistent with this. If we decide to not depend on `Session.getDefaultInstance(System.getProperties(), null).getStreamProvider();` we should carry this decision to the other parts. This means to apply your suggestion in more places where StreamProvider could be null.

Do I implement it?.

jmehrens: Ok. Let's keep default session logic to keep this ticket simple.  I'll do some test to see if default session breaks if implementation loader is shutdown. We can deal with that in a new ticket if needed.

jmehrens (approved):
Your approach is awesome!  Looks good!

jmehrens:
Side note: we'll have to review some of the Angus packages that also extend the classes here. DSN is an example. I glossed over it and I think it is a case of no changes need but something we should keep an eye on.

jmehrens:
@jbescos I wonder if a related PR should be created to force the thread context classloader to loader of the authenticator or Session.class.getClassLoader during the call to Session::getDefaultSession. Or rather when new default session is created.  This would ensure that static default session can pin a StreamProvider in another classloader.  It would also force users to prefer Session::getInstance.  It is a breaking change but for good reason.  Thoughts?

jbescos:
@jmehrens from what I see in the code, when `Session#defaultSession` gets instanced, the `streamProvider` of this `defaultSession` gets instanced too. So, both `defaultSession` and `streamProvider`, come from the same class loader in the moment the `defaultSession` is instanced. Let me know if I missunderstood you or I am wrong.

jmehrens:
@jbescos 

> Let me know if I missunderstood you or I am wrong.

StreamProvider calls [FactoryFinder.find](https://github.com/jakartaee/mail-api/blob/666ec999d8931041d89eabf10d807e5262195f8f/api/src/main/java/jakarta/mail/util/FactoryFinder.java#L38) and that result is what is pinned in the default session.

That code uses the context classloader so in theory default session could pin a webapp classloader when mail-api is the parent loader and implementation is in web app classloader.

My thought behind the proposed new related PR would be to make this edge case fail hard. 
 Consumers would be forced to place mail-api in same loader (webappclassloader) as implementation if they want to use default session. Or consumers must call getInstance inteads.  Otherwise, today webapp classloader A can be pinned and poison default session for webapp in classloader B.

A lot of the behavior depends on the layout of the classloaders and location of api vs implementation. Edge case for sure.  What do you think?

jbescos:
Right, and it is possible that a webapp has its own StreamProvider impl, that will be set in the default session, so all the other wepp apps deployed will use it. I'd really like to remove the static `Session#defaultSession` and let the users to create their own instance and keep it as a singleton if they want.

For the time being, I think it is a good idea what you say. Adding @lukasj in this topic, so he is aware because it is a breaking change.

Issue #777, opened 2025-08-12
services/jakarta.mail.Provider override not working


jbescos:
**Describe the bug**
https://github.com/eclipse-ee4j/angus-mail/issues/170

**To Reproduce**
Attached in the previous issue

jbescos:
@cubistico it was released two days ago: https://mvnrepository.com/artifact/jakarta.mail/jakarta.mail-api/2.1.4

jbescos:
I have updated it, thank you for reminding it to me.

cubistico:
Thank you. However, I'm confused by the fact that this ticket is not mentioned in the release notes?

https://github.com/jakartaee/mail-api/releases isn't update any more? Not insisting, just trying to understand it.

jbescos:
For that one I have updated it here: https://github.com/jakartaee/mail-api/pull/784/files

But I don't know when and how is it going to get deployed there.

I will keep one eye on that, thank you.

PR #779, opened 2025-08-14
services/jakarta.mail.Provider override not working #777

jbescos:
https://github.com/jakartaee/mail-api/issues/777


jmehrens (commented):
This looks great. There's a COMPAT.TXT that should have text added to explain that file names changed and for now there is a fallback to the old names.

jbescos:
I am going to do first some verifications today

Issue #812, opened 2025-09-24
Custom mail providers distributed under META-INF/javamail.properties are no longer loaded with Jakarta Mail 2.1.5


yersan:
**Describe the bug**
Jakarta Mail 2.1.5 is not backward compatible with how custom providers distributed by the application are loaded. Providers declared under `META-INF/javamail.properties` are no longer loaded; users are now forced to rename the file to `META-INF/jakarta.properties` instead. I understand it as a backward compatibility issue.

**To Reproduce**
See reproducer attached to this issue, it is just a test case that loads the session and tries to assert that the custom providers declared on `META-INF/javamail.properties` are available from Session.getProviders()

[mail-reproducer.zip](https://github.com/user-attachments/files/22514615/mail-reproducer.zip)

**Expected behavior**
Custom providers distributed via `META-INF/javamail.properties` are loaded.

**Additional context**
This issue was introduced via [here](https://github.com/jakartaee/mail-api/commit/223c35043429a738320f86abe70c75be1efe7f01#diff-5cbee3986400643f15064c1c111133ea02d864c75ed3ad777d192159f8f6ea89R1285-R1287). As soon as the providers coming from the `META-INF/jakarta.properties` that are available on the classloader via Angus Mail are processed, the for loop is broken and the code skips processing other possible providers distributed by the application via  `META-INF/javamail.properties`

jbescos:
Hi, this is originally the issue related to this topic: https://github.com/eclipse-ee4j/angus-mail/issues/151

If it finds any `jakarta.providers`, it will not process `javamail.providers`, but it is true that in the moment you introduce `jakarta.providers`, any existing user having `javamail.providers` will not be loaded. There is a note [here](https://github.com/jakartaee/mail-api/blob/e041fc2bbc5e0ad66b2ddf7abeab0b1a8566d2ae/www/docs/COMPAT.txt#L23) about this.

There are several ways to address this:
- Depend on `org.eclipse.angus:angus-mail:2.0.3`. This one does not introduce the new jakartified files, so the old `javamail.providers` will be loaded. You don't need to wait for a fix in this option.
- We can load both files (`jakarta.providers` and `javamail.providers`). There will be some duplicates but we can deal with it.
- Load first `javamail.providers`. Not sure this is a good solution, because 2 different apps running in same classpath, one having `javamail.providers` and the other `jakarta.providers`, will exclude the second one.

PR #817, opened 2025-10-01
Custom mail providers distributed under META-INF/javamail.properties are no longer loaded with Jakarta Mail 2.1.5

jbescos:
https://github.com/jakartaee/mail-api/issues/812

[This](https://github.com/jakartaee/mail-api/pull/779/files) could be handy for this review. It contains the changes when only jakarta files were going to be loaded.

Draft for now. I need to check more things.


api/src/main/java/jakarta/mail/Session.java
```diff
@@ -1102,7 +1133,8 @@ private void loadProvidersFromStream(InputStream is) throws IOException {
      */
     public synchronized void addProvider(Provider provider) {
         providers.add(provider);
-        providersByClassName.put(provider.getClassName(), provider);
+        if (!providersByClassName.containsKey(provider.getClassName()))
```
jbescos, line 1136: I was confused here. This is now necessary to avoid jakarta resource overwrite javamail. But why was not this added here before, if we were doing this check in `providersByProtocol` already?.

PR #833, opened 2026-04-21
Handle devanagari email addresses, and add better testing.

### LINKED ISSUE #842: Devanagari not handled in domains
**Describe the bug**

InternetAddress rejects addresses that contain an Indic domain (unless every vowel happens to be an a).

**To Reproduce**
1. Set the allowutf8 property,
2. Parse `"संपर्क@डाटामेल.भारत"` as an InternetAddress.

**Expected behavior**
The domain is valid (both in IDNA2003 and -2008), so I expect parsing to finish without an exception.

**Desktop (please complete the following information):**
 - OS: Linux, with JDK 11 and 21.

**Additional context**
This was discovered while adding units test for [James](https://james.apache.org), forwarded to me, and I traced the problem here. #833 was developed in response to the James problem; this issue exists only so that its number can be mentioned in #833.

arnt:
The code had a stricter idea of valid domain names than IDNA, which broke handling of a few email addresses.

This PR includes unit testing of the address that led to discovering the problem, a few other addresses from the same corpus, a few other addresses from another, and a fix for the problem.

(The root cause is that Devanagari, the script used for e.g. Hindi, doesn't have vowel letters like a/e/i. Rather it has consonants and then modifiers if the vowel isn't a. A word such as sari is written with a "sa" followed by "ra"+"say i instead of a". Those "say i instead" were mishandled because they aren't freestanding letters.)

This solves issue #842.

### DIFF
--- a/api/src/main/java/jakarta/mail/internet/InternetAddress.java
+++ b/api/src/main/java/jakarta/mail/internet/InternetAddress.java
@@ -1385,6 +1385,7 @@ public class InternetAddress extends Address implements Cloneable {
         if (addr.charAt(start) == '.')
             throw new AddressException("Domain starts with dot", addr);
         boolean inliteral = false;
+        boolean idnCheck = false;
         for (i = start; i < len; i++) {
             c = addr.charAt(i);
             if (c == '[') {
@@ -1400,6 +1401,12 @@ public class InternetAddress extends Address implements Cloneable {
             } else if (c <= 040 || c == 0177) {
                 throw new AddressException(
                         "Domain contains control or whitespace", addr);
+            } else if (c == '。' || c == '．' || c == '｡') {
+                // Dot-like code points that java.net.IDN treats as label
+                // separators but which are not RFC 5321 dots. Reject so
+                // the IDN fallback below can't silently accept them.
+                throw new AddressException(
+                        "Domain contains non-ASCII dot", addr);
             } else {
                 // RFC 2822 rule
                 //if (specialsNoDot.indexOf(c) >= 0)
@@ -1415,9 +1422,16 @@ public class InternetAddress extends Address implements Cloneable {
                  * <let-dig> ::= <letter> | <digit>
                  */
                 if (!inliteral) {
-                    if (!(Character.isLetterOrDigit(c) || c == '-' || c == '.'))
-                        throw new AddressException(
-                                "Domain contains illegal character", addr);
+                    if (!(Character.isLetterOrDigit(c) || c == '-' || c == '.')) {
+                        if (c < 0x80)
+                            // ASCII punctuation / specials are genuinely
+                            // illegal in a domain; IDN can't rescue them.
+                            throw new AddressException(
+                                    "Domain contains illegal character", addr);
+                        // Non-ASCII: maybe a combining mark or other IDN
+                        // letter class. Defer to java.net.IDN below.
+                        idnCheck = true;
+                    }
                     if (c == '.' && lastc == '.')
                         throw new AddressException(
                                 "Domain contains dot-dot", addr);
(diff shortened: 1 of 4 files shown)


api/src/main/java/jakarta/mail/internet/InternetAddress.java
```diff
@@ -1427,6 +1441,21 @@ private static void checkAddress(String addr,
...
+            // points like combining marks. Ask java.net.IDN for a second
+            // opinion: if the domain Punycode-encodes cleanly (IDNA2003 +
+            // Nameprep), accept it. The address itself is left byte-for-
+            // byte unchanged.
+            try {
+                java.net.IDN.toASCII(addr.substring(start),
+                        java.net.IDN.ALLOW_UNASSIGNED);
+            } catch (IllegalArgumentException e) {
+                throw new AddressException(
+                        "Domain contains illegal character", addr);
```
jmehrens, line 1456: Maybe chain the IAE to the address exception?

api/src/main/java/jakarta/mail/internet/InternetAddress.java
```diff
@@ -1427,6 +1441,21 @@ private static void checkAddress(String addr,
         }
         if (lastc == '.')
             throw new AddressException("Domain ends with dot", addr);
+        if (idnCheck) {
```
jmehrens, line 1444:
```suggestion
        if (idnCheck && ) allowUtf8
```
If utf8 headers are not enabled I would think we should reject.

jmehrens (changes requested):

api/src/main/java/jakarta/mail/internet/InternetAddress.java
arnt: Maybe I misunderstand — the effect of that seems to be to allow Москва@Москва even without allowUtf8 but disallow भारत@भारत by rejecting the second vertical line after the @.

I can write a separate PR for this file that rejects all non-ASCII, if you want. AFAICT the allowutf8 property was spottily enforced before this change, cleaning tha tup makes sense to me. But it seems orthogonal to the issue in this PR (handling of Indic vowels, which aren't handled like Cyrillic/Greek/… vowels).

arnt: If you'd prefer to have a separate additional commit on this PR that adds thorough testing of allowUtf8, I'll be happy to do it that way too.

api/src/main/java/jakarta/mail/internet/InternetAddress.java
arnt: Good catch, changed and I added another unit test.

api/src/main/java/jakarta/mail/internet/InternetAddress.java
arnt: Actually I'd quite like to write a commit with unit tests that checks for allowUtf8 true/false and such changes as are necessary. I like cleanups. Tell me whether I should use a separate PR or on this, either way is fine IMO.

jmehrens: What you have is fine. We can deal with allow Utf8 flag later.

jmehrens:
Add your ticket number to doc/release/CHANGES.txt
Let me know when you are ready for re-review. Nice work!

arnt:
I filed #842 and updated the code with an extra unit test and the suggested exception change. Ready for new review.

nlisker:
Does this fix #847 too? Is the review stuck somewhere?

arnt:
This PR doesn't contain any changes relating to differences between RFCs 822 and 5322. But I don't know about any related bugs, either, and I know 5322 fairly well... did you have anything particular in mind?

I rebased the PR to the latest HEAD and pushed now.

jmehrens (commented):
I missed it on first review that the copyright end year is 2025 instead of 2026. Please fix and thanks! Looks good otherwise.

nlisker:
> This PR doesn't contain any changes relating to differences between RFCs 822 and 5322. But I don't know about any related bugs, either, and I know 5322 fairly well... did you have anything particular in mind?

Nothing specific. I'm mostly worried about support for non-Latin characters. If those are accepted, and it seems to be from your test cases, then I'm satisfied.

arnt:
> > This PR doesn't contain any changes relating to differences between RFCs 822 and 5322. But I don't know about any related bugs, either, and I know 5322 fairly well... did you have anything particular in mind?
> 
> Nothing specific. I'm mostly worried about support for non-Latin characters. If those are accepted, and it seems to be from your test cases, then I'm satisfied.

Oh! That's in RFCs 6530-3, not 5322, and was already largely supported. This fixes a minor bug.

If there are further problems related to 653x, I'll submit new pull requests.

jmehrens (approved):

jmehrens:
@jbescos I think this looks good. I'll leave it open for a bit and then squash merge it.

arnt:
"A bit", perhaps three weeks? (No hurry, I just have a script that reopens tabs, and ran it today.)

BTW, I think #841 may solve the jenkins failure.

jmehrens:
> "A bit", perhaps three weeks? (No hurry, I just have a script that reopens tabs, and ran it today.)
> 
> BTW, I think #841 may solve the jenkins failure.

One of my reservations for merge was that failure.

I'll set a goal to get this done this week.

jmehrens (approved):

PR #841, opened 2026-04-29
WIP: Use a JDK build with recent TLS.

arnt:
Pull request #833 [failed in CI](https://ci.eclipse.org/mail/job/mail/job/PR-833/2/stages/); the root cause appears to be that the JDK used in Jenkinsfile is many years old and its TLS doesn't interoperate well with that of post-2024 maven repos. This switches to a recent Temurin build, which seems to work better (but I've marked this WIP so CI won't spam maintainers if I'm wrong).

As a bonus, it tells the CI to use UTF-8, which is important to many of the new tests in #833. It seems to me that the pom.xml setting should be enough, but what do I know.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -70,11 +70,10 @@ spec:
       steps {
         container('mail-ci') {
           sh """
-            wget 'https://download.java.net/openjdk/jdk11/ri/openjdk-11+28_linux-x64_bin.tar.gz' -O openjdk-11.tar.gz -q
-            tar -xzf openjdk-11.tar.gz
-            cd jdk-11
-            export JAVA_HOME=`pwd`
-            cd ..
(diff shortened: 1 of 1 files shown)


arnt:
I see that a maintainer needs to bless this to even run the CI for testing. Sorry. Dear maintainers, will you please try? Thanks.

pzygielo:
> I see that a maintainer needs to bless this to even run the CI for testing.

But that's needed for github workflows. I'm pretty sure modified Jenkins is not used for that.

Perhaps

> https://ci.eclipse.org/mail/job/mail/view/change-requests/

is indexing on schedule only. Time will tell.

pzygielo:
And, I'm guessing, this modification would not be considered as trusted by Jenkins instance (and the version from target branch (w/o fix) would be used instead) anyway. So some intervention by maintainers might be needed to verify this change works and does not break anything else.

arnt:
I'll close this now.

I still am not 100% certain why CI is so unhappy about #833. When I looked at it, this PR seemed likely to solve the problem. But three months have passed, it's time to move on.

